PR #230, opened 2016-02-12
Add Google Code World educational environment. It is a nice start for…

ner0x652:
… those interested in learning programming using a FP programming language


JackMorganNZ:
Hi @ner0x652, this looks like a great resource but we are actually removing this list altogether from the introduction. We believe there are plenty of programming resources of there, with plenty of sites listing these so it saves us having to maintain a list of sites ourselves.

If we do keep the list in a hidden teacher panel, we may add this resource to the list.

Thanks for the suggestion.

PR #232, opened 2016-02-13
Remove unnecessary parenthesis. For a reader that does not know Pytho…

ner0x652:
…n, I think, it adds useless details(symbols)

### DIFF
--- a/files/linear-binary-search-python2.py
+++ b/files/linear-binary-search-python2.py
@@ -9,7 +9,7 @@ import time
 from random import randint
 
 def binary_search(array, item):
-    #Performs a Binary search and returns the number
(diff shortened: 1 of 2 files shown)


Heidirose:
I agree completely with removing those parenthesis. It isn't even standard style as far as I know.

Thanks for the contribution!

JackMorganNZ:
Thanks @ner0x652, I'll merge these changes in. 

In the future, it would be great to pull from the `develop` branch instead. I still need to write the contributing documentation to explain this though.

JackMorganNZ:
We have now switched the default branch on GitHub to be `develop` which should solve any PR issues.

PR #239, opened 2016-02-19
readability improvement for linear search python 3 version. Output ex…

ner0x652:
…ecution time in miliseconds as in linear search python 2 version


JackMorganNZ:
Great change, thanks @ner0x652

Issue #242, opened 2016-02-23
Background color difference in HCI


ner0x652:

ner0x652:
![Uploading Screenshot from 2016-02-23 21:15:24.png…]()
![Uploading Screenshot from 2016-02-23 21:15:24.png…]()

ner0x652:
![screenshot from 2016-02-23 21 15 24](https://cloud.githubusercontent.com/assets/2598963/13263646/1c09fd7a-da73-11e5-8200-0ce413a912b0.png)

JackMorganNZ:
The highlight is because the URL will have a permalink.

If you click on `Interface Usability` on the side navigation on the right, it will move to that section, without creating a permalink.

If you click on the heading `4.3 Interface Usability` then your URL will change to http://www.csfieldguide.org.nz/en/chapters/human-computer-interaction.html#interface-usability (notice the permalink at the end).

If you refresh the page or click the link with a permalink, the heading will be highlighted (though the image you attached is highlighting the whole section as the link was http://www.csfieldguide.org.nz/en/chapters/human-computer-interaction.html#section-interface-usability). 

We added the highlight because it's very useful when following a glossary link to see where it's been linked ([try this example](http://www.csfieldguide.org.nz/en/chapters/formal-languages.html#glossary-chomsky-hierarchy-2)), however the example you showed isn't ideal. Two solutions:

1. The highlight fades over several seconds.
2. The highlight isn't used for permalinks starting with `section-`.

I prefer the second option, thoughts?

JackMorganNZ:
I changed the JS code to only highlight sections if they didn't start with `section-`, to avoid whole sections being highlighted.

PR #243, opened 2016-02-23
change to solve #242. background-color property to transparent

ner0x652:


JackMorganNZ:
Hi @ner0x652, see my comment in #242, this code needs to remain as is but there are other solutions I can easily implement.

PR #270, opened 2016-03-21
Fix small issues in Graphics chapter

StevenMaude:
Fix some formatting and text typos.

Also fix a couple of issues with the Bresenham line solution, and
clarify behaviour of the 2D rotation matrix.


JackMorganNZ:
Great changes and documented perfectly!

PR #293, opened 2016-04-30
Fix #134: Center images

### LINKED ISSUE #134: Images in some chapters not correctly centering
Have a look at the HCI chapter, at the scroll wheels and the estimation comic (in the "Visibility of system status" section.

oughter:
Changes:
* Make images display as `inline-block` instead of `block`, so they center properly in their parent div.

To test:

1. Go to output/en/chapters/human-computer-interaction.html#visibility-of-system-status
1. Note the progress indicator images (and all others) are now properly horizontally centered
1. View the same page in mobile layout and note that the images are now centered

### DIFF
--- a/generator/scss/third-party/components/_global.scss
+++ b/generator/scss/third-party/components/_global.scss
@@ -129,6 +129,7 @@ i {
 // Images
 img.responsive-img,
 video.responsive-video {
+  display: inline-block;
   max-width: 100%;
   height: auto;
 }


Heidirose:
Thanks! That bug had been irritating me :-)

JackMorganNZ:
Hi @oughter, thanks for the suggestion. We have also tried to make the images centered as well (just realised it was never logged on #134) but we encountered an issue with [MaterializeCSS's Material Box](http://materializecss.com/media.html), sadly this fix hits the same issue.

The image is centered to the user on the page, but when the image is selected to zoom in the image cuts back to the original position and zooms in/out from there. It looks quite jarring so we left it as is.

Though we also have issues with Material Box zooming images larger than their original size which creates distorted images.

It may be best for us at the moment do drop Material Box completely at the moment and center all images, while we investigate better options for allow users to view the details of an image.

I should be able to remove Material Box later this week (unless you wish to complete it earlier). I'll leave this PR open for the moment.

Thanks, Jack.

oughter:
@JackMorganNZ, thanks for the details on the images issues. Stopping use of Material Box for now due to these issues sounds like a good approach.

I'll take a stab at removing Material Box this week but will let you know if something comes up and I'm unable to get to it.

JackMorganNZ:
Solved by PR #300.

PR #300, opened 2016-05-03
Replace image lightbox implementation

oughter:
Material Box lightboxing of images is problematic (see #293 for details), so I replaced its use with a different library: [Featherlight](https://noelboss.github.io/featherlight/).

Reasons for choosing Featherlight:
* It has an MIT license (should not be any issues with UCCSER using it)
* It's open source and is actively maintained
* It seems to be fairly customizable and easy to work with

I tried to replicate the existing experience provided by Material Box. One difference is the growing/shrinking effect when an image is zoomed in/out that MB provides is not available with Featherlight (at least not easily achievable). MB is able to do that easily because it uses the same image for the normal view and zoomed view; however, that is related to the positioning problems noted in #293.

Another difference in the Featherlight implementation is that images are never scaled up past the image's full size. I personally prefer this, but that restriction could be overridden with some CSS changes if desired.


JackMorganNZ:
Thanks @oughter, this is great work! I love the work you have done. I've pulled the branch into our repo and made some changes:

- Correctly loaded `featherlight.min.js`
- Added white background to images with transparent backgrounds (quite a few in the Programming Languages chapter)
- Added zoom out icon to whole overlay
- Move the custom CSS rules to `website.scss`. This keeps all our custom CSS rules in the one file, which is evaluated by the SCSS processer. End result is the same, but easier to manage our code from MaterializeCSS.
- I think I got images to display at 100% size (rather than 94%) without scroll bars (`overflow: hidden` on the parent did the trick).
- Updated documentation.

I've been checking images throughout the guide and worked well so I've merged in into out development branch. Thanks again!

I have one question for you: We are currently still writing our documentation, but even without it, how hard was it to figure out the structure of the repo (especially dealing with setting files and HTML templates)?

oughter:
@JackMorganNZ, thanks for going over my work!

> We are currently still writing our documentation, but even without it, how hard was it to figure out the structure of the repo (especially dealing with setting files and HTML templates)?

I didn’t have trouble with the repo structure, and I was able to quickly trace through the generate python script to figure out where files get loaded from (although I apparently didn’t do the third-party JS script loading correctly). As someone who recently worked on the codebase without prior exposure to it, I can provide feedback for the documentation if that will be helpful.

PR #321, opened 2016-05-08
Interactive caesar

Jamesernator:
This adds README.md and cleans up some code to make it easier to understand.


interactives/caesar-cipher/index.html
```diff
@@ -14,6 +14,7 @@
 </head>
 
 <body>
+    <div class="interactive row">
```
JackMorganNZ: @Jamesernator There is a duplicate `div` here with no end. This line should probably be removed.

PR #325, opened 2016-05-09
Content/data representation

Heidirose:
This branch also has @Jamesernator's interactives that were made for the data representation chapter in it. Is that okay?


text/en/chapters/data-representation.md
```diff
@@ -880,11 +880,9 @@ In order to compare and evaluate them, we need to decide what it means for a rep
 We know that UTF-8, UTF-16, and UTF-32 can represent all characters, but ASCII can only represent English. Therefore, ASCII fails the first criterion.
 But for the second criteria, it isn't so simple.
 
-The following interactive will allow you to find out the length of pieces of text using UTF-8, UTF-16, or UTF-32. Find some samples of English text and Asian text (forums or a translation site are a good place to look), and see how long your various samples are when encoded with each of the three representations.
+The following interactive will allow you to find out the length of pieces of text using UTF-8, UTF-16, or UTF-32. Find some samples of English text and Asian text (forums or a translation site are a good place to look), and see how long your various samples are when encoded with each of the three representations. Copy paste or type text into the box.
 
-{comment}
-Insert interactive here which determines the length of a piece of text with each representation
-{comment end}
+{interactive name="unicode-length" type="iframe"}
```
JackMorganNZ: @Heidirose @Jamesernator Is there any reason this is included as an `iframe` rather than `in-page`?

Heidirose: The reason I did it is because in the past I've found that somewhat similar interactives have similar variable names in them, and it can collide and make a mess, unless we use an iframe. I used an iframe as a precaution, although perhaps it isn't necessary.

I'm not sure if the collision issue still exists in CSFG

Heidirose:
Note that this has some incomplete work in it (the colours guide).

It is not actually linked into the guide though, so should be okay

JackMorganNZ:
@Heidirose Thanks for letting me know, for the sake of sanity we will leave it there. I'm going to do one last comb through and check that everyone works as I've fixed most of the bugs, but some fixes may have broken other content.

PR #332, opened 2016-05-11
Interactive/high score boxes

thevictorchang:
Includes HCI menu too. high-score-boxes interactive teaches searching algorithms using boxes
that reveal a number on click. Boxes are vertically shifted to create an unordered look.


JackMorganNZ:
I'll review this sometime this week.

JackMorganNZ:
Hi Victor,

I've had a look at both the action menu interactive, here are my comments:

**Action Menu Interactive**

- [ ] The JS file should be within a folder called `JS`
- [ ] The link to the JS should be within the div with the `interactive` class so it is properly copied.
- [ ] It would be good if the user could click another option after one is selected (unless this is against the point of the interactive), so instead of removing the element you could display the text via `alert();`
- [ ] The `actionmenu.css` file is unused and doesn't seem to be required so can be deleted.
- [ ] `README.md` file is missing.

I'll look at the other interactive early next week.

thevictorchang:
Hi Jack

I have made the required changes to the file structure and added a
README.md.

Allowing the user to click another option after one is selected is indeed
against the point of the interactive - demonstrating the HCI concept of
"not putting complex/important capabilities in a non complex/important
interface".

Thanks

Victor

interactives/action-menu/index.html
```diff
@@ -0,0 +1,40 @@
...
+    <link rel="stylesheet" href="../base-files/css/third-party/materialize.min.css" type="text/css" media="screen,projection"/>
+    <script src="../base-files/js/third-party/modernizr.js"></script>
+    <!-- jQuery (necessary for Materialize) -->
+    <script type="text/javascript" src="../base-files/js/third-party/jquery.js"></script>
+  </head>
+  <body>
+    <div class="row interactive center-align">
+    <div class="s12 col">
+    <!-- ### Start of required HTML for inline interactives ### -->
+      <a id='dropdownMenu' class='dropdown-button btn' href='#' data-activates='dropdown-available-menu-items' data-beloworigin='true'>Available Menu Items</a>
```
JackMorganNZ: It would be good for both interactives to use more precise selectors, that won't interfere with other interactives. You don't need to worry about this if the interactive is being used in `whole-page` mode (which I believe the `high-score-boxes` is designed for this), however `in-page` interactives should be more precisely specified. I can show you this when I see you next.

interactives/high-score-boxes/css/style.css
```diff
@@ -0,0 +1,59 @@
+
+body {
+}
+
+/* covering boxes */
+div.box {
+    width: 100px;
+    height: 100px;
+    /*border: 10px;*/
```
JackMorganNZ: It would be nice to delete unused lines from files (unless it may be used in the future).

interactives/high-score-boxes/css/style.css
```diff
@@ -0,0 +1,59 @@
...
+	position: relative;
+	/*padding: 50px, 50px, 50px, 50px;*/
+    margin-top: 40px;
+    margin-bottom: 40px;
+    margin-right: 20px;
+    margin-left: 20px;
+}
+
+#restartButton {
+	display: none;
```
JackMorganNZ: Might be nice to add some padding to the button so it's slightly larger.

interactives/high-score-boxes/index.html
```diff
@@ -0,0 +1,48 @@
...
+			</div>
+		</div>
+		<div class="col s12">
+			<form name="myForm"  action="congratulations.html"
+			onsubmit="return validateForm()" method="post" >
+            <input type="text" id="answerTextForm" placeholder="Enter answer here..." name="submittedInt">
+			<input type="submit"  value="Submit">
+			</form>
+            <p id="completionMessage"> </p>
+            <a id="restartButton" class="btn btn-large btn-expand" >Restart</a>
```
JackMorganNZ: This button is not rendering correctly as it's being displayed `inline` rather than `block`. Try changing the element from an `a` element to a `button` element. Also the `btn-expand` class is doing nothing.

JackMorganNZ:
We should also add the interactives into the chapters as well so users can see them.

thevictorchang:
Hi Jack

I have made the necessary changes and git pushed. I will come by your
office sometime this week and we can discuss the JQuery selectors and
adding the interactives into the chapters.

Thanks

Victor

JackMorganNZ:
Excellent! Those changes look great, I'll merge this in for the `v2.6` release.

interactives/high-score-boxes/js/findHighScore.js
```diff
@@ -0,0 +1,180 @@
...
+	for (var i = 0; i < (boxes.length); i++) { 
+		if (document.getElementById('box' + i) ==  event.target) {
+			$(this).fadeOut(1000);
+			$(this).fadeIn(1000);
+			boxes[i].revealed_times += 1;
+		}
+	}
+	//to start timer only if it hasn't started yet
+	if (!gameStarted) {
+		for (var i = boxes.length - 1; i >= 0; i--) {
```
JackMorganNZ: Can you tell me what lines 48 and 49 are for? They seem like they are not needed.

PR #404, opened 2017-02-10
Handle digits in 2d-arrow-manipulation config name

Goldob:
There is a bug in `2d-arrow-manipulations` interactive that occurs when using config names including digits, e.g. `matrix-scale-translate-2`. It is caused by a faulty regex that matches incomplete label and makes the script look for a non-existent config file.

This PR fixes that.


JackMorganNZ:
Thanks @Goldob, looks great at first glance. Will check it properly and merge if all working when we return to work next week. 😄

hayleyavw:
Thanks for that! :)

PR #470, opened 2017-07-30
Interactive - JPEG Compression

isabelle49:
An interactive that allows the user to understand how JPEG compression works.
Stepper library is used to allow dragging of numbers under the basis functions.


JackMorganNZ:
Looks awesome. I'm currently in the process of merging this in, there is a few changes I'll make before it's merged in:
- Change the `button` tag to the `interactive` tag as it has a bunch of extra features and checks.
- Add a thumbnail image.
- Some small HTML changes so it wraps better on small screens:
  image

On a side note, I get this error sometime, I think from moving my mouse over the interface:

```
interact.min.js:32
Uncaught TypeError: a[Oa] is not a function
    at R (interact.min.js:32)
    at d (interact.min.js:43)
    at Array.B.forEachSelector (interact.min.js:97)
    at xa.pointerOver (interact.min.js:44)
    at HTMLDocument.<anonymous> (interact.min.js:16)
```

isabelle49:
Thanks for that! I'm assuming you changed the first few things, but I can look into that strange error and how to make the layout better on small screens. Also I can disable the image dragging, definitely seems unnecessary. Do I follow the same process of making the changes on a different branch and creating a pull request?

JackMorganNZ:
Yep, new pull request as before. I'll let you make the rest of the changes. Thanks!

PR #510, opened 2017-10-18
Fix bug in calculating UTF-16 encoding (fixes #501)

### LINKED ISSUE #501: Check UTF encoding interactives
**_This issue is user submitted and needs to be validated:_**

When the letters "abcd" are put into the Unicode length text box, both UTF-8 and UTF-16 are shown to use 32 bits. However, if UTF-16 uses 16 bits to encode a character, this should be 16*4=64 bits. Is there a problem with this or is this intentional?

**_Also reported by another user:_**

In the data representation section, the Unicode Encoding Size interactive has a bug where the same number of bits is displayed for UTF-16 as it is for UTF-8, allowing for values such as 8 or 24 bits as the length for a piece of text for UTF-16. This is very misinforming and I hope it is fixed soon so that students do not put this error into their work. Thanks :)

**_Also reported by another user:_**

I think there is a problem with the interactive that lets you compare text samples to see how many bits they will use depending on which UTF encoding scheme you use. It appears to give the same results for both UTF-8 and UTF-16. It would be great if you could check this out as students will be keen to use it for their upcoming NCEA external assessments in Computer Science.
http://csfieldguide.org.nz/en/chapters/data-representation.html#comparison-of-text-representations

**_Also reported by another user:_**

I have come across what I think is a bug on your the Unicode Encoding Size interactive at the bottom of the Unicode section. 

It is always showing the same size for UTF-8 and UTF-16. For example, the letter 'a' is saying UTF-8 size = 8 (correct) and UTF-16 = 8 (which is not possible. It should be 16)
Another example is the character '猫'. It says that both UTF-8 and UTF-16 is equal to 24 whereas it should be UTF-8=24 and UTF-16 = 16

Can you please let me know if this is indeed a bug or if I have missed something.

JackMorganNZ:
This pull request fixes the Unicode Length interactive to correctly determine the length of UTF-16 interactives. 

It also:
- Reduces the file size for the user as third party packages are not used.
- Avoids compiling CoffeeScript into JavaScript by using native JavaScript.


interactives/base-files/js/unicode/unicode.js
```diff
@@ -0,0 +1,25 @@
+function bitLength(text, utf_encoding) {
+  // Returns the bit length of an string for the given UTF encoding
```
jordangriffiths01: s/an/a

interactives/unicode-length/js/unicode-length.js
```diff
@@ -0,0 +1,8 @@
+$(document).ready(function () {
+  $("#interactive-unicode-length-button").on('click', function() {
+    var text = $("#interactive-unicode-length-text").val();
+    $("#interactive-unicode-length-utf8").html(bitLength(text, 8) + " bits");
+    $("#interactive-unicode-length-utf16").html(bitLength(text, 16) + " bits");
+    $("#interactive-unicode-length-utf32").html(text.length * 32 + " bits");
```
jordangriffiths01: Perhaps this should be included in the unicode.js bitLength function too? I know it's trivial, but that way the logic is kept together. 

Not too fussed either way, I'll leave it up to you :)

JackMorganNZ: This is specific to this interactive, while `bitLength` was separated out as I'll import it for other Unicode interactives.

PR #514, opened 2017-10-19
Compressed all imagens on the project

kauelima:
Compressed up to 10mb on images


JackMorganNZ:
Hi @kauelima,

Thanks for spending the time on this pull request, however sadly we are currently rewriting the entire project system, and part of this will be a pipeline for image manipulation. We will also be resizing images where appropriate, and then compressing them after this.

So I will close this issue until this pipeline is completed, however if we ever need these images in the future we can then merge it in.

Thanks for the time you spent doing this :smile: 

Cheers,

Jack Morgan and the CSFG team

PR #522, opened 2017-11-13
Django basic structure

hayleyavw:
Ok there is a lot going on here. Think of this PR as the starting point for forming the new Django backend for CSFG.

the tl;dr; of this PR is basically it puts the system at a point where we should be able to plan and prioritise tasks to get it up and running on a dev deployment and eventually at production level as it should now be a lot easier for multiple people to work on it and to tackle individual pieces of the puzzle. 

This PR includes:
- General application
- Chapters application
- Interactives application
- Settings for django, docker, travis, etc

These are really basic applications to get us started. There are basic models/urls/views/loaders set up to load chapters and interactives (although interactives are currently hardcoded). Right now the system is able to take a chapter (markdown file) and convert it using Verto and display it on a page that looks very similar to the current CSFG site (the Introduction and Algorithms chapters have been converted to Verto syntax for testing).

Interactives: inline interactives can be displayed without issue, whole page interactives can load their css and js files, but images are a different story because of the way django stores static files, open to suggestions on how to deal with this issue. iframe interactives are yet to be implemented.

The current settings used to generate and run the system are more or less copy/pasted from CSU as at mid July, i.e. will be changing dramatically over the coming weeks. What is there now was set up purely just to get the thing running.

There are tests, but they are not to be trusted......


csfieldguide/templates/base.html
```diff
@@ -18,99 +18,82 @@
     <link rel="stylesheet" href="{% static 'css/website.css' %}">
     <link rel="stylesheet" href="{% static 'css/third-party/featherlight.min.css' %}">
     <link rel="shortcut icon" href="{% static 'img/guide-favicon.png' %}">
-    <script src="{% static 'js/third-party/modernizr.js' %}"></script>
+    <script type="text/javascript" src="{% static 'js/third-party/modernizr.js' %}"></script>
+    <script type="text/javascript" src="{% static 'js/third-party/jquery.js' %}"></script>
```
JackMorganNZ: jQuery should be loaded at the end of the page, not within `<head>`. There is some great explanations why here: https://stackoverflow.com/questions/2105327/should-jquery-code-go-in-header-or-footer

csfieldguide/utils/custom_converter_templates/interactive.html
```diff
@@ -0,0 +1,22 @@
+{%- if type == 'in-page' -%}
+<remove>
+{{ "{% include 'interactives/" }}{{ name }}{{ ".html' %}" }}
+</remove>
```
JackMorganNZ: What are these `<remove>` tags for?

csfieldguide/templates/interactives/prototype-whole-page-interactive-base.html
```diff
@@ -0,0 +1,31 @@
+{% load static %}
```
JackMorganNZ: Should `csfieldguide/templates/interactives/prototype-whole-page-interactive-base.html` still be included?

csfieldguide/utils/custom_converter_templates/interactive.html
hayleyavw: no idea, most of these templates bits were copied from the original system

csfieldguide/templates/base.html
hayleyavw: yes I agree, but it needs to be loaded first in order for inline interactives to work. This seems terrible and like something that can be fixed, but not in a one line change that I know off the top of my head :(

csfieldguide/utils/custom_converter_templates/interactive.html
hayleyavw: have `remove`d them (see what I did there?.... 😉 😆 ) and replaced with a `div`

PR #524, opened 2017-11-14
Issue/492

hayleyavw:
Now uses the base docker images from the docker-images repo.

These settings from the csu repo with adjustments and deletions where necessary. Would be helpful if reviewers could thoroughly check the changes to spot any residual csu references and to make suggestions of edits for the csfg context, which will turn into a list of issues to fix/add over time.


csfieldguide/gulpfile.js
```diff
@@ -0,0 +1,201 @@
...
+var postcss = require('gulp-postcss');
+var postcssFlexbugFixes = require('postcss-flexbugs-fixes');
+var autoprefixer = require('autoprefixer');
+var sourcemaps = require('gulp-sourcemaps');
+
+// linting
+var jshint = require('gulp-jshint');
+var stylish = require('jshint-stylish');
+
+// Scratch image rendering
```
JackMorganNZ: The Scratch image rendering packages and logic can be removed as it's unused in CSFG.

Issue #541, opened 2017-12-15
Roboto font and Material Design icons cannot be found


hayleyavw:
This interferes with page load

PR #543, opened 2017-12-15
Issue/542

hayleyavw:
Added icons for each chapter.

Also fixed issue where materialize fonts could not be found, dramatically improving loading times. (#541)
Also playing with a new layout for the chapters index page (if suggested layout is accepted, an issue will need to be created for replacing temporary text for chapter descriptions).


csfieldguide/static/css/style.css
```diff
@@ -0,0 +1,15 @@
+img.chapter-icon {
+	color: red;
```
JackMorganNZ: This doesn't appear to be used.

csfieldguide/static/css/style.css
```diff
@@ -0,0 +1,15 @@
+img.chapter-icon {
```
JackMorganNZ: These icons would look better if the lines were purple on a white background.

csfieldguide/static/css/style.css
```diff
@@ -0,0 +1,15 @@
...
+	color: red;
+	background-color: #7280ce;
+	width: 20% !important;
+	border: 5px #7280ce solid;
+	border-radius: 4px;
+}
+
+.chapter-list {
+	padding-top: 20px;
```
JackMorganNZ: This would be better stored as `margin-top: 1em` to match paragraph spacing.

csfieldguide/templates/chapters/index.html
```diff
@@ -13,15 +13,25 @@ <h1 id="chapters" class="section-heading anchor-link">Chapters</h1>
...
+    
+      {% if all_chapters %}
+          {% for chapter in all_chapters %}
+          <div class="l12 chapter-list-item">
+            <div class="row">
+              <a href="{% url 'chapters:chapter' chapter.slug %}">
+                <img class="col l3 chapter-icon" src="{% if chapter.icon %}{% static chapter.icon %}{% else %}http://placehold.it/238x200{% endif %}">
+                <div class="col l9">
+                  <h3>{{ chapter.name }}</h3>
+                  <p>This is example text. This is where a short description of the chapter will go. This is example text. This is where a short description of the chapter will go. This is example text. This is where a short description of the chapter will go.</p>
```
JackMorganNZ: Can we delete this until implemented?

csfieldguide/templates/chapters/index.html
```diff
@@ -13,15 +13,25 @@ <h1 id="chapters" class="section-heading anchor-link">Chapters</h1>
...
-        </a>
-      {% endfor %}
-  {% else %}
-    <p>No chapters are available.</p>
-  {% endif %}
+  <div class="row chapter-list">
+    
+      {% if all_chapters %}
+          {% for chapter in all_chapters %}
+          <div class="l12 chapter-list-item">
```
JackMorganNZ: `chapter-list-item` has no CSS rules attached to it so should be deleted.

csfieldguide/templates/chapters/index.html
```diff
@@ -13,15 +13,25 @@ <h1 id="chapters" class="section-heading anchor-link">Chapters</h1>
...
+            <div class="row">
+              <a href="{% url 'chapters:chapter' chapter.slug %}">
+                <img class="col l3 chapter-icon" src="{% if chapter.icon %}{% static chapter.icon %}{% else %}http://placehold.it/238x200{% endif %}">
```
JackMorganNZ: Using `l3` without `s12` seems to work, but it might be good to define mobile sizing to declare correct width CSS rule.

csfieldguide/templates/chapters/index.html
```diff
@@ -13,15 +13,25 @@ <h1 id="chapters" class="section-heading anchor-link">Chapters</h1>
...
-      {% endfor %}
-  {% else %}
-    <p>No chapters are available.</p>
-  {% endif %}
+  <div class="row chapter-list">
+    
+      {% if all_chapters %}
+          {% for chapter in all_chapters %}
+          <div class="l12 chapter-list-item">
+            <div class="row">
```
JackMorganNZ: It would be nice if all this content is center aligned when viewing on a mobile device.

csfieldguide/templates/chapters/index.html
JackMorganNZ: Would be better to show no image than a placeholder, so could use:
```
{% if chapter.icon %}
    <img class="col l3 chapter-icon" src="{% static chapter.icon %}">
{% endif %}
```

csfieldguide/templates/chapters/index.html
hayleyavw: see lines 13-15 in `style.css`

JackMorganNZ: That could be simplified as:

```
 .chapter-list p {
  color: #000;
}
```

csfieldguide/templates/chapters/index.html
```diff
@@ -13,15 +13,24 @@ <h1 id="chapters" class="section-heading anchor-link">Chapters</h1>
...
-    <p>No chapters are available.</p>
-  {% endif %}
+  <div class="row chapter-list">
+
+      {% if all_chapters %}
+          {% for chapter in all_chapters %}
+          <div class="l12">
+            <div class="row">
+              <a href="{% url 'chapters:chapter' chapter.slug %}">
+                <img class="col l3 chapter-icon" src="{% if chapter.icon %}{% static chapter.icon %}{% endif %}">
```
JackMorganNZ: The sizing of this icon isn't working correctly because the `l3` class is setting the width to 25% of the container, while the `chapter-icon` at 20% (which seems to be 20% of the row but due to negative margins it is actually bigger than the `l3` class). To solve this problem, the image should be wrapped in the column class like so:

```html
<div class="col l3">
  <img class="chapter-icon responsive-img" src="{% if chapter.icon %}{% static chapter.icon %}{% endif %}">
</div>
```

And the `width: 20% !important;` should be removed from the `chapter-icon` class. The new class `responsive-img` sets the image to always be 100% of the parent.

csfieldguide/static/css/style.css
```diff
@@ -0,0 +1,15 @@
+img.chapter-icon {
+  fill: red;
```
JackMorganNZ: I can't see where this is used?

localhost_8081_en-gb_chapters_

hayleyavw: oh sorry, keep accidentally leaving in my tests

csfieldguide/templates/chapters/index.html
hayleyavw: thanks

PR #558, opened 2018-01-23
Fix computed value

anzeljg:
Computed value is not correctly shown since it only shows the power of
base in certain place. For example when using calculator with base 10 it
shows only 1, 10, 100, 1000 etc. instead of seleted value in drop-down
multiplied with either 1, 10, 100, 1000 etc. depending on the place in the
number.


JackMorganNZ:
Hi @anzeljg, thanks so much for spotting this bug and thanks again for taking the time to fix it! I'll be sure to list your contribution on the 'Contributors' page in the next release.

anzeljg:
No problem :)
Glad I could help improving this fantastic resource.

Issue #453, opened 2017-06-14
Check for converter required files


hayleyavw:

eAlasdair:
completed

PR #569, opened 2018-02-13
Issue/562 - breaking up chapters into sections

hayleyavw:
In the interest of getting the review process started, am creating this PR having only broken up one of the chapters currently in the new system. The remaining 3 chapters will also need to be broken down into sections before this pr is merged.

This PR contains:
- Adapted introduction chapter, which is now split into multiple sections
- Adapted `Chapter` model
- New `ChapterSection` model
- New `ChapterSectionLoader`
- Adapted urls and views
- Adapted and new templates for displaying chapters and using correct links
- Tests for all of the above

@JackMorganNZ things in particular I would like input on:
- Layout of the first section in the introduction chapter, particularly, what is it's heading meant to be (or does it even need one)?
- Currently there is only one configuration file, but I am thinking this should be broken down into multiple config files, one for the application, and another for each chapter to list it's sections. Thoughts?
- Does the way I have created the models make sense, or is there an alternative way you think it should be done?


csfieldguide/chapters/views.py
```diff
@@ -23,6 +24,17 @@ def get_queryset(self):
         """
         return Chapter.objects.order_by("number")
 
+    def get_context_data(self, **kwargs):
+        context = super(IndexView, self).get_context_data(**kwargs)
+        chapter_sections = ChapterSection.objects.all()
+        for chapter in self.object_list:
+            chapter.first_section = chapter_sections.get(
```
JackMorganNZ: What is this accomplishing?

csfieldguide/general/context_processors/nav_context.py
```diff
@@ -1,6 +1,6 @@
 """Context processor for displaying chapters in nav bar."""
```
JackMorganNZ: I feel like this can be accomplished not as a context processor, therefore not being run on every page.

csfieldguide/general/context_processors/nav_context.py
```diff
@@ -9,4 +9,12 @@ def nav_context(context):
     Returns:
         Dictionary containing list of chapter objects
     """
-    return {'chapters': Chapter.objects.all()}
+    chapters = Chapter.objects.all()
+    # Add first chapter section to each chapter
+    chapter_sections = ChapterSection.objects.all()
+    for chapter in chapters:
```
JackMorganNZ: This seems really inefficient, as it's getting the first item for every chapter, just to have one chapter's title. Unless I'm missing the point of this completely?

csfieldguide/templates/chapters/chapter_section.html
```diff
@@ -0,0 +1,31 @@
...
+  {% render_html_field chapter_section.content %}
+
+  <div id="chapter-section-buttons">
+    {% if previous_section %}
+      <a href="{% url 'chapters:chapter_section' chapter.slug previous_section.0.slug %}" class="btn-menu">
+        Previous: {{ previous_section.0.heading }}
+      </a>
+    {% endif %}
+    {% if next_section %}
+      <a href="{% url 'chapters:chapter_section' chapter.slug next_section.0.slug %}" class="btn-menu">
```
JackMorganNZ: This button should always be right aligned.

csfieldguide/chapters/management/commands/_ChaptersLoader.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @transaction.atomic
+    def load(self):
+        """Load the content for a chapter.
+
+        Raises:
+            MissingRequiredFieldError: When a config (yaml) file is missing a required
+                field.
+        """
+        chapter_number = self.chapter_structure.get("chapter-number", None)
```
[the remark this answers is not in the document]
hayleyavw: following the convention we set in csu where we are always pretty explicit, did check csu for the current convention (incase something had changed) before doing it here.

hayleyavw:
@JackMorganNZ This is ready for re-review.

Couple of things:
- I cannot seem to right align that damn `Next: section` button to save my life, input is very welcome
- Checks for required files is not yet implement in csfg (and is in issue #453 )

csfieldguide/chapters/management/commands/_ChaptersLoader.py
```diff
@@ -34,51 +37,62 @@ def load(self):
...
+
+        sections_yaml = self.chapter_structure.get("sections", None)
+        if sections_yaml is None:
             raise MissingRequiredFieldError(
-                self.structure_file_path,
-                ["sections", "chapter-number", "title"],
-                "Chapter sections"
+                self.chapter_structure_file_path,
+                self.required_fields,
+                "Chapter"
```
JackMorganNZ: I believe this is the wrong term here?

csfieldguide/chapters/management/commands/loadchapters.py
```diff
@@ -49,10 +48,25 @@ def handle(self, *args, **options):
...
-                chapter_structure,
-                BASE_PATH
-            ).load()
+        else:
+            for chapter_slug in chapters:
+                chapter_structure_file_path = os.path.join(
+                    BASE_PATH,
+                    "{}/{}.yaml".format(chapter_slug, chapter_slug)
+                )
+                chapter_number = chapters[chapter_slug]["chapter-number"]
```
JackMorganNZ: Wouldn't this logic be better in the chapter loader?

csfieldguide/chapters/models.py
```diff
@@ -32,8 +33,9 @@ class Chapter(models.Model):
     slug = models.SlugField(unique=True)
     name = models.CharField(max_length=100)
     number = models.SmallIntegerField(unique=True)
+    introduction = models.TextField()
     other_resources = models.TextField(null=True)
```
JackMorganNZ: What is `other-resources` for a chapter? This seems to be a duplicate of the "Further information" section of a chapter. I don't think this attribute is required for a chapter but keen to hear your thoughts?

csfieldguide/chapters/models.py
```diff
@@ -71,6 +73,24 @@ def __str__(self):
         """
         return self.heading
 
+    def clean(self):
```
JackMorganNZ: Awesome to see this, especially as it leverages Django (I learnt new stuff!)

csfieldguide/chapters/management/commands/_ChaptersLoader.py
hayleyavw: It looks right to me...
This throws the following error when a chapter yaml file is missing the `sections` key.
```
A Chapter requires the following field:
  - ['sections', 'icon']

For the missing field:
  - Is the field name spelt correctly?
  - Does the field have the correct value?
```

csfieldguide/chapters/management/commands/loadchapters.py
hayleyavw: The chapter number is specified in the `structure.yaml` file (along with the chapter slug). Therefore it makes sense to have this code in this file because this is where `structure.yaml` is parsed.
I could move this to the chapter loader, but the number would still have to be extracted in loadchapters anyway, and additionally, the `structure.yaml` file path would have to be given as an additional parameter in order to be used in the error handling code for checking the chapter number is valid.
Overall, it seemed much cleaner to have this code in the loadchapters file.
Thoughts?

csfieldguide/chapters/models.py
hayleyavw: there is actually an even more awesome way to do it, using validators, but ran into quite a bit of trouble trying to implement it so this was a good plan B, just not quite as cool. (both are as good as each other to use/I don't think there is a good practice bias, I did a bit of reading).

hayleyavw:
Re chapter icon, see my previous comment
> Checks for required files is not yet implement in csfg (and is in issue #453 )

hayleyavw:
Re aligning the buttons, float was the very first thing I tried and it's just not working.

csfieldguide/chapters/management/commands/loadchapters.py
JackMorganNZ: My thoughts were thinking of where it would be loaded irrelevant to the source files. I thought of the logic in three parts:

1) Module for handling Django command call.
2) Module for loading chapters.
3) Module for loading chapter sections.

If I were to guess where the logic for the chapter number is located without looking at files, it would be in number 2.

Would it be better to move a topic's data to its own configuration file (like CS Unplugged's topics) so extra values like authors can be easily added later? Or would you rather have it as one longer file?

JackMorganNZ:
Do you want to commit your `float` CSS and we can go from there?

hayleyavw:
bump @JackMorganNZ

csfieldguide/chapters/management/commands/loadchapters.py
JackMorganNZ: @hayleyavw Pointed out in offline discussion that listing all chapter numbers together makes it easy for authors to see/modify order of all chapters. So happy to leave this as is.

csfieldguide/chapters/urls.py
```diff
@@ -24,11 +24,16 @@
         views.glossary_json,
         name="glossary_json"
     ),
-    # eg: /chapters/algorithms/
+    # eg: /chapters/introduction/
```
JackMorganNZ: Would be great to change this example to URL and the one on line 33 to a chapter that isn't labelled 'introduction' as I initially thought this was referring to the introduction attribute of the model. Something like 'algorithms' would avoid that short confusion.

csfieldguide/chapters/views.py
```diff
@@ -23,6 +23,21 @@ def get_queryset(self):
         """
         return Chapter.objects.order_by("number")
 
+    def get_context_data(self, **kwargs):
```
JackMorganNZ: This method isn't required (see comment on template).

csfieldguide/chapters/views.py
```diff
@@ -40,11 +55,79 @@ def get_queryset(self):
...
-    """View for a specific topic."""
+    """View for intro section of a chapter."""
 
     model = Chapter
-    template_name = "chapters/chapter.html"
-    slug_url_kwarg = "chapter_slug"
+    template_name = "chapters/chapter_index.html"
+    context_object_name = "chapter"
+
+    def get_object(self, **kwargs):
```
JackMorganNZ: Is this custom method required? Would the default method suffice?

csfieldguide/templates/chapters/index.html
```diff
@@ -14,12 +14,11 @@ <h1 id="chapters" class="section-heading anchor-link">Chapters</h1>
 {% block content %}
 
   <div class="row chapter-list">
-
       {% if all_chapters %}
           {% for chapter in all_chapters %}
           <div class="l12">
             <div class="row">
-              <a href="{% url 'chapters:chapter' chapter.slug %}">
+              <a href="{% url 'chapters:chapter_section' chapter.slug chapter.first_section.slug %}">
```
JackMorganNZ: Pretty sure the original code here is correct, as a link from the chapter index will jump straight to the first section, missing the introduction!

Issue #622, opened 2018-05-01
Update interactives for Django system


JackMorganNZ:
- [x] 2d-arrow-manipulations (**Renamed 2d-shape-manipulations** (see #863))
- [x] action-menu
- [x] available-menu-items
- [x] awful-calculator
- [x] base-calculator
- [x] big-number-calculator @courtneycb
- [x] binary-cards @courtneycb
- [x] bin-packing
- [x] box-rotation
- [x] box-translation
- [x] caesar-cipher **coffeescript**
- [x] checksum-calculator @courtneycb
- [x] checksum-calculator-gtin-13 @courtneycb
- [x] close-window @courtneycb
- [x] cmy-mixer @courtneycb
- [x] colour-matcher @courtneycb
- [x] compression-comparer @courtneycb
- [x] confused-buttons @courtneycb
- [x] date-picker
- [x] deceiver @courtneycb
- [x] delay-analyser @courtneycb
- [x] delayed-checkbox @courtneycb
- [x] frequency-analysis @courtneycb **coffeescript**
- [x] fsa-box @courtneycb
- [x] fsa-light @courtneycb
- [x] fsa-washing-machine @courtneycb
- [x] hex-background-colour @courtneycb
- [x] high-score-boxes
- [x] huffman-tree @courtneycb
- [x] image-bit-comparer @courtneycb
- [x] jpeg-compression
- [x] MD5-hash @courtneycb (**Deleted in #984**)
- [x] mips-assembler
- [x] mips-simulator
- [x] ncea-guide-selector (**Deleted in #958**)
- [x] nfa-guesser
- [x] no-help @courtneycb 
- [x] number-generator @courtneycb
- [x] packet-attack
- [x] packet-attack-level-creator
- [x] parity @courtneycb
- [x] payment-interface @courtneycb
- [x] pixel-viewer
- [x] python-interpreter
- [x] regular-expression-filter
- [x] regular-expression-search
- [x] rgb-mixer @courtneycb
- [x] rsa-jsencrypt
- [x] rsa-key-generator 
- [x] rsa-no-padding
- [x] run-length-encoding
- [x] searching-algorithms
- [x] sha2 **coffeescript**
- [x] sorting-algorithm-comparison @courtneycb
- [x] sorting-algorithms
- [x] starter-structure (not doing)
- [x] trainsylvania @courtneycb
- [x] trainsylvania-planner @courtneycb
- [x] trig-function-calculator
[...]

PR #632, opened 2018-06-05
No help refactor

courtneycb:
Refactor the no-help interactive to use the Django web framework. Relates to #622.


csfieldguide/static/interactives/no-help/scss/no-help.scss
```diff
@@ -0,0 +1,6 @@
+#interactive-no-help {
+    background-color: #4caf50;
```
JackMorganNZ: Rather than set the background, border and text colours, you can use `btn-success` instead of `btn-primary` to get the green button with white text.
See: https://getbootstrap.com/docs/4.1/components/buttons/

csfieldguide/static/interactives/no-help/scss/no-help.scss
```diff
@@ -0,0 +1,6 @@
...
+    border-color: #4caf50;
+    color: #fff;
+    margin-bottom: 14px;
```
JackMorganNZ: Instead of specifying an exact margin, you can use Bootstrap margin helpers to add the spacing.
See: https://getbootstrap.com/docs/4.1/utilities/spacing/
They are very useful once you start using them. In this case I would remove this style rule and add `mb-3` onto the `a` element.

csfieldguide/templates/interactives/no-help.html
```diff
@@ -0,0 +1,19 @@
+{% extends interactive_mode_template %}
+
+{% load static %}
+
+{% block html %}
+    <div class="row interactive center-align">
+        <div class="col s12">
+            <a role="button" class="btn btn-primary btn-lg" id="interactive-no-help">Click me for help</a>
```
JackMorganNZ: I forgot to say that all text needs to be translatable. The page on Django about this is here: https://docs.djangoproject.com/en/1.11/topics/i18n/translation/

This is a big topic, so happy to talk in person about this.

csfieldguide/templates/interactives/no-help.html
```diff
@@ -0,0 +1,19 @@
+{% extends interactive_mode_template %}
+
+{% load static %}
+
+{% block html %}
+    <div class="row interactive center-align">
```
JackMorganNZ: You won't need to specify a row and column here, a simple center aligned `div` will work, or using auto horizontal margins (see: https://getbootstrap.com/docs/4.1/utilities/spacing/#horizontal-centering).

JackMorganNZ: The `interactive` class is not required anymore in the new system.

csfieldguide/templates/interactives/no-help.html
courtneycb: I think I managed to do this in the templates file but might need to chat to you about translation in JavaScript files

csfieldguide/static/interactives/no-help/js/no-help.js
```diff
@@ -0,0 +1,3 @@
+$("#interactive-no-help").click(function(){
+    alert("Error: No help available!")
```
JackMorganNZ: Sorry, this text needs to be translated. Can discuss it tomorrow.

PR #635, opened 2018-06-07
Delayed checkbox refactor

courtneycb:
Refactor the delayed-checkbox interactive to use the Django web framework. Relates to #622.


csfieldguide/templates/interactives/delayed-checkbox.html
```diff
@@ -0,0 +1,22 @@
...
+
+{% load i18n %}
+{% load static %}
+
+{% block html %}
+  <div id="interactive-delayed-checkbox">
+    <h4>{% trans "Tick the checkbox" %}</h4>
+    <input id="interactive-delayed-checkbox-box" type="checkbox" value="">
+      <div id="interactive-delayed-checkbox-feedback">
+        {% trans "Clicked " %}<span id="interactive-delayed-checkbox-counter">{% trans "0" %}</span>{% trans " times" %}
```
JackMorganNZ: This line should be translated all at once, since different languages may display the order differently. The HTML is fine within the translation tags, and you won't need to translate "0" as it's number.

PR #642, opened 2018-06-13
Refactor trainsylvania-planner interactive.

courtneycb:
Refactor the trainsylvania-planner interactive to use the Django web framework. Relates to #622.


csfieldguide/static/interactives/trainsylvania-planner/css/trainsylvania-planner.css
```diff
@@ -1,24 +1,42 @@
 #interactive-trainsylvania-planner-input {
-  font-size: 1.2em;
-  width: 80%;
-  text-transform: uppercase;
+    background-color: transparent;
+    border: none;
+    border-bottom: 1px solid #9e9e9e;
```
JackMorganNZ: Sorry for not saying this earlier but the input boxes don't need to match the previous MaterializeCSS style, then can be Bootstrap default (the previous style is also hard to decipher what it is).

PR #644, opened 2018-06-19
Unicode length refactor

courtneycb:
Refactor the unicode-length interactive to use the Django web framework. Relates to #622.


JackMorganNZ:
Could you update me where the functions `utf16BitsUsed` and `utf8BitsUsed` are sourced from.

PR #654, opened 2018-06-29
Refactor delay-analyser interactive.

courtneycb:
Refactor the delay-analyser interactive to use the Django web framework. Relates to #622.


JackMorganNZ:
@courtneycb I've updated this interactive to not require the Foundation framework, would be good if you could test it. I deleted the Foundation files from this PR, but there is a few other interactives that will require them when we get around to converting them. I'll try to convert these too, otherwise we can re-add the files.

PR #660, opened 2018-07-02
Refactor parity interactive.

courtneycb:
Refactor the parity interactive to use the Django web framework. Relates to #622.


courtneycb:
There's something weird going on with the css file diff.. Also, the green success border on the cards doesn't clear when the reset button is clicked.. this is how it functions in prod though so I didn't change anything as I thought it might be like that for a reason.

JackMorganNZ:
@courtneycb Defintely was a bug, so have pushed a simple fix.

PR #662, opened 2018-07-02
Add search feature

JackMorganNZ:
Adds search feature to CSFG.

If testing locally, requires rebuild of Docker images.


csfieldguide/templates/base.html
```diff
@@ -46,6 +48,20 @@
                 {% trans "Appendices" %}
               </a>
             </div>
+            <form method="get" action="{% url 'search:index' %}">
+              <div class="input-group" id="search-navbar">
+                <input type="search" class="form-control form-control-sm border-0" placeholder="Search" aria-label="Search" name="q" required="required">
```
courtneycb: I never thought of this on my own work, but can placeholders be translated? / Should they be?

PR #679, opened 2018-07-16
Refactor rsa-no-padding interactive.

courtneycb:
Refactor the rsa-no-padding to use the Django web framework. Relates to #622.


csfieldguide/templates/interactives/rsa-no-padding.html
```diff
@@ -0,0 +1,47 @@
+{% extends interactive_mode_template %}
+
+{% load i18n %}
+{% load static %}
+
+{% block html %}
+	<div class="container">
+  	<div class="row">
```
JackMorganNZ: I think this repeated `row` and `col-12` elements, plus the `container` element are not needed. Might be worth checking.

PR #713, opened 2018-09-13
Use Django's Javascript translation catalog

courtneycb:
This PR enables the use of Django's [JavaScript translation catalog](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#internationalization-in-javascript-code), meaning we can now use gettext() to translate text in JavaScript files.

It also uses [django-statici18n](https://django-statici18n.readthedocs.io/en/latest/index.html), a Django app that collects JavaScript catalogs into a single location. (Instead of generating it dynamically on each request).

**Todo:**
- [x] Documentation


csfieldguide/config/urls.py
```diff
@@ -16,6 +17,7 @@
     url(r"^chapters/", include("chapters.urls", namespace="chapters")),
     url(r"^appendices/", include("appendices.urls", namespace="appendices")),
     url(r"^search/", include("search.urls", namespace="search")),
+    url(r"^jsi18n/$", JavaScriptCatalog.as_view(), name="javascript-catalog"),
```
JackMorganNZ: This might not be required, worth checking.

PR #715, opened 2018-09-13
Add spacing between interactives and text below for all in-page interactives.

courtneycb:


JackMorganNZ:
Do you think this should be applied to all `in-page` interactives?

JackMorganNZ:
@courtneycb Tests need updating (but more likely changing altogether as it shouldn't be checking CSS classes).

PR #734, opened 2018-09-26
Fix delay-analyser interactive

courtneycb:


csfieldguide/static/interactives/delay-analyser/js/delay-analyser.js
```diff
@@ -1,7 +1,7 @@
 // Delays stored in milliseconds
-this.delay_types = [0, 25, 50, 75, 100, 150, 200, 350, 600];
-this.gridSize = 6; // Multiples of three required
-this.total_delays_for_each = (gridSize * gridSize) / delay_types.length;
+var delay_types = [0, 25, 50, 75, 100, 150, 200, 350, 600];
+var gridSize = 6; // Multiples of three required
+var total_delays_for_each = (gridSize * gridSize) / delay_types.length;
```
JackMorganNZ: Should we do something to these variables to ensure another in-page interactive doesn't use these variables? (like put them into an object)

Or does JS already scope by file?

Do you want to research this?

courtneycb: Good point. I'll put them into an object since JS doesn't scope by file.

courtneycb: They can actually be made local. Also tidied up the inconsistent naming style.

Issue #732, opened 2018-09-25
HCI chapter contains images that use the English language


courtneycb:
The images in the following doc need to be translated or hidden when viewed in a language other than English.

https://docs.google.com/document/d/1Sk_emztUzfOGwBAbX9aMng7vzw-Y_x3t6GyxDG0SLkU/edit?usp=sharing

PR #740, opened 2018-09-28
Add confusing-error interactive to replace image in HCI chapter

courtneycb:
Replace
image

with
image

So that the words can be translated. Relates to #732


csfieldguide/templates/interactives/confusing-error.html
```diff
@@ -0,0 +1,23 @@
...
+{% load i18n %}
+{% load static %}
+
+{% block html %}
+	<div class="container">
+		<div class="row justify-content-center">
+		  <div class="card col-lg-8 col-xs-12">
+		  	<div class="card-body text-left">
+		  		<h6 class="card-title">{% trans "This configuration will cause your backup device's status light to blink amber. Are you sure you want to continue?" %}</h6>
+		  		<img class="float-left icon-4 mr-2" src="{% static 'interactives/confusing-error/img/icons8-data-backup-100.png' %}">
```
JackMorganNZ: Because we are using an icons8 image, we need to add attribution in the footer. Could you please check what is required and add it? I have done it here: https://github.com/uccser/ktam-content-assistant/blob/develop/templates/base.html#L136 but best to check that's correct.

PR #774, opened 2018-10-24
Consistent numbering for chapter sections

### LINKED ISSUE #773: Numbers inconsistent for chapter sections
It appears that if a section number is skipped in the structure files, there are inconsistencies in the user interface.

## Example

*In particular, see the numbering for Sorting section*

image

image

## Possible Solutions

The first image uses a numbered list to show the sections, so if section 1 doesn't exist, section 2 is labelled as 1. To solve this we could use the section number attribute (2.2) from the model with the chapter's number attribute (as the full number gives more context).

However there is still a problem with this, as the list could contain:

```
2.1
2.2
2.4
2.7
```

This would leave the user confused to where the other numbers are. A solution for this is to make the list up to the highest available number, and if there is not a section for the number then display the text "Section currently available".

However it doesn't make sense for there to be gaps in the English section numbers (people would question what is this missing section and when it would become available) so the loader should check if the numbers are sequential, and throw an exception if they are not available. If a translation (non-English language) doesn't have that section available, it can display the English title with the "Not available in language XYZ" badge.

courtneycb:
Resolves #773 

Expands on #773 by:

- showing 'not-available-warning' when a chapter or chapter section is not translated in the current language.

- Greying out chapter section links if they are not translated in the current language

### DIFF
--- a/csfieldguide/templates/chapters/chapter.html
+++ b/csfieldguide/templates/chapters/chapter.html
@@ -30,6 +30,9 @@
         <a href="{% url 'chapters:chapter_section' chapter.slug section.slug %}">
           {{ section.name }}
         </a>
+        {% if not chapter_section.translation_available %}
+          {% include "generic/not-available-badge.html" %}
+        {% endif %}
       </li>
     {% endfor %}
   </ol>
--- a/csfieldguide/templates/chapters/chapter.html
+++ b/csfieldguide/templates/chapters/chapter.html
@@ -12,29 +12,37 @@
 {% endblock page_heading %}
 
 {% block content %}
-  {% if chapter.video %}
-    <div class="row mb-3">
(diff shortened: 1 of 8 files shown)


csfieldguide/templates/chapters/chapter.html
```diff
@@ -12,26 +12,37 @@
...
   {% endif %}
-  {% render_html_field chapter.introduction %}
 
   <h2>Sections in this chapter:</h2>
-  <ol>
+  <ul class="list-unstyled">
     {% for section in chapter_sections %}
       <li>
-        <a href="{% url 'chapters:chapter_section' chapter.slug section.slug %}">
+        {{ chapter.number }}.{{ section.number }}
```
JackMorganNZ: For consistency with other parts of the website, the number always ends in a full stop.
```suggestion
        {{ chapter.number }}.{{ section.number }}.
```

PR #794, opened 2018-11-06
Add test for checking all 'package.json' files are included in root 'package.json'

JackMorganNZ:
This PR adds a basic test to check if all files are listed in the root `package.json`.


csfieldguide/tests/test_repository.py
```diff
@@ -0,0 +1,38 @@
...
+class RepositoryTests(SimpleTestCase):
+    """Test class for tests for other areas of the project repository."""
+
+    def test_package_json_includes_all_modules(self):
+        """Check if all 'package.json' files are listed in the root 'package.json' file."""
+        ROOT_PACKAGE_JSON = "package.json"
+        PATH_PREFIX = "file:./"
+        IGNORE_FILES = [
+            "static/interactives/caesar-cipher/package.json",
+            "static/interactives/unicode-binary/package.json",
```
hayleyavw: what's special about these two?

JackMorganNZ: These are not included as they currently require a module that doesn't exist anymore (was a bug @courtneycb and I fixed on Monday). The `package.json` files are still there so we can fix them later, but these files are not listed in the root `package.json` file so the build can complete. The JavaScript files that are trying to use the modules are completely commented out.

Issue #783, opened 2018-11-01
Address feedback form design suggestions


hayleycodes:
- [ ] External links could open in a new tab (multiple people have asked for this)
- [x] Side bar should be able to be hidden (multiple people have asked for this)

eAlasdair:
As of 3.0.0 the side bar no longer exists

aenkirch:
I think I resolved the issue of external links (see #1175)

PR #806, opened 2018-11-14
Change external html links to open in a new tab

eAlasdair:
Addresses `External links could open in a new tab (multiple people have asked for this)` from #783 by adding `target="_blank"` to all external links


courtneycb:
Need to update the links that were added in from merging develop and add documentation.

eAlasdair:
Before this is merged, the notes by courtneycb should be implemented and an update to Verto should be made to allow the same behaviour from external links written in Markdown files

PR #817, opened 2018-11-29
Refactor Bin Packing Interactive

eAlasdair:
- Refactor the Bin Packing Interactive to use the Django framework
- Allow translations
- Implement a check to prevent bad size values being used to create boxes
Relates to #622 

bin-packing


csfieldguide/static/js/interact.min.js
```diff
@@ -0,0 +1,7 @@
+/* interact.js v1.4.0-alpha.17+sha.793e5fa-dirty | https://raw.github.com/taye/interact.js/master/LICENSE */
```
JackMorganNZ: This file shouldn't be required as it's loaded from NPM.

csfieldguide/templates/interactives/bin-packing.html
```diff
@@ -0,0 +1,58 @@
...
+  </div>
+{% endblock html %}
+
+{% block css %}
+  <link rel="stylesheet" href="{% static 'interactives/bin-packing/css/bin-packing.css' %}">
+{% endblock css %}
+
+{% block js %}
+  <script type="text/javascript" src="{% static 'interactives/bin-packing/js/bin-packing.js' %}"></script>
+  <script type="text/javascript" src="{% static 'js/interact.min.js' %}"></script>
```
JackMorganNZ: This line shouldn't be required if `interact.js` is included in the main JS file.

PR #860, opened 2019-02-04
Prevent back of binary cards showing through once flipped

### LINKED ISSUE #859: Back of binary cards showing through on Firefox browsers
Issue when displaying flipped cards in Firefox browser:

image

To solve this issue the following CSS rule needs to added:

```css
.binary-card {
    backface-visibility: hidden;
}
```

courtneycb:
Resolves #859


csfieldguide/static/interactives/binary-cards/css/binary-cards.css
```diff
@@ -67,3 +67,6 @@
 h3, h5, p {
   text-align: left;
 }
+.binary-card {
```
JackMorganNZ: There is already a rule for this (line 16), it would be nicer to have the rule in there.

PR #868, opened 2019-02-11
Fix interactive thumbnails not displaying on dev site

### LINKED ISSUE #858: Interactive thumbnails not deployed
Currently all interactive thumbnails are not showing: https://cs-field-guide-dev.appspot.com/en/interactives/

eAlasdair:
Fixes #858 where files were not properly migrated before thumbnails could be created


csfg
```diff
@@ -96,6 +97,21 @@ cmd_update() {
 }
 defhelp update 'Run Django migrate and updatedata commands and build static files.'
 
+# Update the system for production deployment
+cmd_update_prod() {
```
JackMorganNZ: I think this function is still a bit unclear. It reads like it's the `update` command but with production static files. But then it's also missing components that are in the `update` that aren't used for static files. If I went `./csfg start` and `./csfg update_prod` I would not get a working website.

I feel like this function should be exactly the same as `update` with `static` as `static_prod`. Or the script for making production files needs to be in the deployment script only (which is what it's purpose is).

Thoughts?

eAlasdair: I must've made a mistake testing then because it worked for me. Okay perhaps taking the update then redeploy method you mentioned yesterday would be best; the wasted time is minimal and doesn't happen very often after all

PR #869, opened 2019-02-11
Allow the loading of iframes if they are from the same domain

### LINKED ISSUE #867: X-Frame-Options set to DENY in HTTP response header
This disallows us from loading our interactives/resources in an iframe.

We need to set the X-Frame-Options setting to sameorigin. This will allow interactives/resources to be loaded that come from within our domain.

courtneycb:
Resolves #867


infrastructure/nginx/nginx.conf
```diff
@@ -27,5 +27,7 @@ http {
           proxy_redirect off;
           proxy_pass http://django:8080;
         }
+
+        add_header X-Frame-Options sameorigin;
```
JackMorganNZ: This is only used for the local dev environment, has not effect to Google App Engine.

PR #886, opened 2019-02-26
Fix No3 of Thumbnails not deployed on dev site

eAlasdair:
Relates to #858 
I'll explain why I believe this will work by explaining an issue I noticed last year:

- `./csfg update_static` deletes all the thumbnails without recreating them.
- It executes two commands: `./csfg static` and `./csfg collect_static`.
- I had thought this wasn't the issue, because `./csfg update` runs both those commands yet the thumbnails are properly deployed in testing.
- I've now realised the command `./csfg make_interactive_thumbnails` is run after `./csfg static` but before `./csfg collect_static` in this update command.
- So the theory is: `./csfg static(& static_prod)` deletes the thumbnails, and neither `./csfg update_static` nor the release build commands remake the thumbnails afterwards.

It won't fix everything if the server problem mentioned in #858 is a separate issue, but just adding `./csfg make_interactive_thumbnails` back in should solve the issue at least partially.


courtneycb:
Resolves #858

Issue #451, opened 2017-06-13
List third party licences


hayleyavw:
In `LICENCE-THIRD-PARTY` and in `third-party-licences/`

eAlasdair:
#863 will introduce the directory and begin this process

eAlasdair:
All thrid-party licences have now been added

PR #903, opened 2019-04-04
Activate NetCom Protocols Chapter

### LINKED ISSUE #118: Fix formatting + check content for Network Protocols

eAlasdair:
This PR activates the Network Communication Protocols chapter

resolves #118

## Checks

- [x] All related interactives work properly:
  - [x] Packet Attack
  - [x] Packet Attack Level Creator
- [x] Full internationalization support
- [x] Properly documented

## Changes:

- [x] Content fixes
- [x] Create uninteractives for images with text for internationalisation
  - [x] `packet-structure-diagram.png` (`tcp-segment`)
  - [x] `udp-encapsulation.png` (`udp-layers`)
- [x] Fix Packet Attack interactive (Relates to #622)
  - Update to Phaser 3
  - Reimplement original functionality (see notes)
  - Introduce better Phaser style and helpful classes
  - Introduce kill animation
  - Introduce replay level button
  - Full internationalisation support
- [x] Fix Packet Attack Level Creator interactive (Relates to #622)
- [x] Introduce proper licence files (Relates to #451)
- [x] Remove now unused images and files

## Notes:

- Reviewers please check that the instructions for viewing Network data in the Google Chrome dev tools are still correct - I only have firefox installed
- Note that there are known issues listed in the Packet Attack README
- Packet Attack scoring functionality has not been reimplemented.
- Packet Attack uses the font `Open Sans` from Google Fonts. This *must* be loaded before the game is loaded, or text will be invisible. As such the very start screen has been moved out of the game, and no game assets will be loaded until the user presses 'start'. This is a rather inefficient solution, so in case a better solution is found in future, the underlying code for this screen has been left in.
- Packet Attack Level Creator does not have the option to allow ACKs/NACKs to be vulnerable. This is the same as its previous iteration.
- The replacement uninteractive `tcp-segment` recreates the original image, but could be a misleading representation of packet structure. Header sections that should be twice the size of others are displayed as the same size. The text describing the image has been updated with this in mind, but perhaps the uninteractive should be changed to better reflect it
- The paragraph about a bus as an example of packet congestion (in the TLP section) is poor in my opinion. In reality buses are used to reduce congestion, and the solution proposed avoids the problem by increasing the number of pathways instead of fixing it in the original scope of the problem. I have proposed a replacement paragraph that better represents this, and gotten Tim's approval.


csfieldguide/chapters/content/en/network-communication-protocols/sections/transport-layer-protocols.md
```diff
@@ -8,23 +8,39 @@ These protocols take different approaches to ensure data is delivered in an effe
...
-The road is jammed and there’s no way you and your friends are getting home on time.
-So you decide to get out of the bus and go your own separate ways.
+In short, the website looks at the image and breaks it into packets.
+It then sends the packets over the internet to your computer.
+Your computer then reassembles the packets to give back to you as the image.
+This is [demonstrated in this video](https://www.youtube.com/watch?v=ewrBalT_eBM).
+
+By now you’re probably wondering why we bother splitting up packets… wouldn’t it be easier to send the file as a whole?
+Well, it solves congestion.
+Imagine you're returning a book you borrowed from your mate on the other side of the railway, and just before you get to the crossing a great long train begins lumbering through it.
```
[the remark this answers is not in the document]
eAlasdair: I'm going to change most of it, from Tim's feedback

csfieldguide/templates/interactives/packet-attack.html
```diff
@@ -0,0 +1,40 @@
+{% extends interactive_mode_template %}
+
+{% load i18n %}
+{% load static %}
+
+{% block html %}
+  <div class="my-3 container-fluid">
```
courtneycb: Are we able to center this?

csfieldguide/static/interactives/packet-attack/js/game.js
```diff
@@ -0,0 +1,901 @@
...
+        scene.levelText.setText(TXT_LEVEL + ' ' + scene.levelNum);
+        scene.levelMessage = CONFIG.LEVELS[level].message;
+        scene.sendText.setText(TXT_SENDING + '\n' + scene.levelMessage);
+    }
+
+    /**
+     * Handler function for a 'receivedMessage' registry update.
+     * Updates the UI to present the given message
+     */
+    updateReceivedMessage(scene, message) {
```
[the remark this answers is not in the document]
eAlasdair: The game ends immediately after, so it's never visually apparent. I'll add a delay

courtneycb:
The links in the udp-layers uninteractive are unclickable because of the css class `.uninteractive` in `website.scss`. It declares` pointer-events` to `none`. This basically means that the element doesn't exist to cursors and you can click right through it. You will have to override it for those links.

csfieldguide/chapters/content/en/network-communication-protocols/sections/application-level-protocols.md
```diff
@@ -81,59 +87,63 @@ A URI is a unique code or number for a resource.
...
+Near the top you’ll see a request to an `html` file.
 Click that and you’ll see details of the Headers, Preview, Response, Cookies and Timing.
 Ignore those last two for now.
 
 Let’s look at the first few lines of the headers:
 
 ```
 Remote Address: 132.181.2.122:3128
-Request URL: http://www.csfieldguide.org.nz/en/index.html
+Request URL: http://www.csfieldguide.org.nz/en/
```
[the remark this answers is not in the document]
eAlasdair: Could remove `en/` entirely? It redirects to the version with a language

csfieldguide/templates/interactives/packet-attack.html
eAlasdair: No, I've tried a lot of things but the canvas decides it must be left-aligned

eAlasdair:
This is not yet ready to merge

courtneycb:
Instructions for viewing Network data in the Google Chrome dev tools have been checked and are fine.

csfieldguide/chapters/content/en/network-communication-protocols/sections/transport-layer-protocols.md
```diff
@@ -31,12 +31,12 @@ So each page is split up and sent in tiny pieces, then at the other end they are
 
 {panel type="curiosity"}
 
-# Three days?!
+# A full day?!
 
-At 86 terabytes of code, Google's database is one of the largest in the world.
-To write that all down and store in 40ft shipping containers, would create a train as long as the distance from Barcelona to Moscow!
+According to one [2016 estimate](https://www.gearthblog.com/blog/archives/2016/04/big-google-earth-database.html), Google Earth has approximately 3 petabytes (3000000Gb) of past and present image data on our world from above.
```
courtneycb: The b in "Gb" needs to be capitalised as a lowercase b means bits and I think you want bytes.

csfieldguide/chapters/content/en/network-communication-protocols/sections/transport-layer-protocols.md
```diff
@@ -31,12 +31,12 @@ So each page is split up and sent in tiny pieces, then at the other end they are
...
+To write that all down and store in 40ft shipping containers would create a train over a thousand kilometers long &ndash; as long as the distance from Auckland to Dunedin!
```
courtneycb: Should we stick to one measurement system? Maybe change 40ft to 12m?

eAlasdair: I'm using '40ft shipping container' to refer to the standard rather than the length. I think most metric system users would understand
https://en.wikipedia.org/wiki/Intermodal_container
I can change it if you want, but I can't say 12m (I'd need to use 12.19) because that's significant enough to make me redo all the calculations

courtneycb: What does the calculation come to if you use 12m? You've used assumptions elsewhere in the calculation so I'm sure rounding it is fine as you're not giving an exact result anyway. I think it's neater to keep the same units especially as you should work in the same units when doing caluclations anyway.

eAlasdair: Okay I'll recalculate it and get back to you

csfieldguide/chapters/content/en/network-communication-protocols/sections/application-level-protocols.md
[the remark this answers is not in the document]
courtneycb: Yeah do that

csfieldguide/chapters/content/en/network-communication-protocols/sections/transport-layer-protocols.md
eAlasdair: I recalculated both and the difference isn't too significant, however I don't like using containers that don't quite fit the standard so I'm keeping the actual measurements from the 40ft version. As for what to write, I reworded it a little so the `12m` is in there, but if you insist I'll remove the 40ft part

courtneycb: I'm fine with both in there

PR #930, opened 2019-05-04
Do not check default interactive thumbnail paths

JackMorganNZ:
- Removes tests were static files were searched for, as these are not used.
- The loaders still check if the interactives are available.
- Remove all unused manual thumbnails


JackMorganNZ:
Tests are failing due to some jobs (different each time) failing to download `django-haystack`.

```
Collecting django-haystack==2.8.1 (from -r /requirements/base.txt (line 17))
  Downloading https://files.pythonhosted.org/packages/69/43/3e247b7b2134b48e9a53fb387e191e5e05b5f38f2faf78ca892097c2b441/django-haystack-2.8.1.tar.gz (1.6MB)
```

Will restart them occasionally, as it appears to be an issue out of our control.

JackMorganNZ:
Tests now passed, ready for review.

JackMorganNZ:
> Will need to do one of: leave it in; rename it and change how its referenced (as far as I remember there's just one reference to the file); or remove it and change the error note about it in the HTML

Thanks, I'll definitely rename it to make it clearer for its intended purpose.

PR #932, opened 2019-05-07
Change default button colour for bootstrap primary buttons.

courtneycb:
The predefined colour for bootstrap buttons with the `btn-primary` class is blue. This style of button is used heavily throughout csfg and doesn't really fit with our colour theme. This PR changes the `btn-primary` colour to that of csfg-light.


eAlasdair:
> Everything looks good, though the thumbnails don't update the first time `./csfg update` is run and I do think that might turn into an issue when it goes live

Actually no, `update` is run before `make_interactive_thumbnails` because of the way I structured the production build. But if that gets changed there might be an issue

PR #952, opened 2019-05-14
Make iframe backgrounds transparent

### LINKED ISSUE #829: Make iframe interactives transparent
Currently iframe interactives look like this:
screen shot 2018-12-17 at 1 25 48 pm

We should make the background transparent so they appear like this:
screen shot 2018-12-17 at 1 26 34 pm

eAlasdair:
Resolves #829 by adding a new stylesheet specifically for iframe content, and adding a rule to it that specifies 'body' have a transparent background.

image

All content in an iframe is affected. This could have unintended side effects if other elements in individual cases rely on the background being white. The transparent background will be unapparent if elements on top of the 'body' (i.e. in the interactives themselves) have their own background colour set.

**I do not know if the project style/structure is correct,** could you check please @JackMorganNZ?


csfieldguide/templates/interactives/base/iframe.html
```diff
@@ -23,6 +23,7 @@
     {% block css %}
     {% endblock css %}
 
+    <link rel="stylesheet" href="{% static 'css/interactive-iframe.css' %}">
```
JackMorganNZ: I think (and it's worth checking) to move this line above line 23 so the rules are defined before any  interactive style rules, allowing the interactive to override them. Need to checkup on precedence of external style rules.

eAlasdair: I put it there because it mirrors how the JS line (~30) is done relative to the JS block. I'll look up what order they're applied

eAlasdair: You are correct @JackMorganNZ, the style rule needs to be before the css block to allow it to be overridden so I changed the order. However while testing this I found something interesting:
image
Overriding the `<body>` element in an interactive that's only found in an iframe, unlike this stylesheet here, overrides everything, including body elements outside the iframe. More rules could be added in future of course but as it stands there's no good reason to edit the body element for an interactive anyway

JackMorganNZ: That doesn't seem right at all, could you send me an example?

eAlasdair: No my mistake, both ones I tested had in-page alternatives in the chapter

JackMorganNZ: Okay, but still confused on how this image exists since the stylesheet it limited to within iframes. Or did you create this with another method?

eAlasdair: I added a <body> style in the interactive css which overrode the background colour. The blue background in the interactive is from that particular interactive (i.e. it overrode the transparent colour), the blue background of the actual page is from the same interactive used in-page further up (i.e. it overrode the white colour it normally is).

JackMorganNZ: Okay, but now we don't use that approach?

eAlasdair: Which approach? Because it's an `in page` interactive further up it has css rules that apply everywhere. So the takeaway is; don't override css rules that are used by the main page content. That approach hasn't changed, and there's no need to edit the `body` element anyway.

> I think (and it's worth checking) to move this line above line 23 so the rules are defined before any interactive style rules, allowing the interactive to override them. Need to checkup on precedence of external style rules.

What you asked me to do is to check if interactives are allowed to override the rules defined in that file. I have ensured that they are but also confirmed that it is a bad idea for them to do so. This is consistent with the other interactive types

JackMorganNZ: The approach of editing a interactive's CSS, and as you say, we don't edit it due to weird side effects (especially when modified the `body` element).

I think what confused me was misreading:

> I added a <body> style in the interactive css which overrode the background colour.

as meaning the new iframe CSS file, but you are meaning a specific interactive CSS file.  Makes more sense once I've re-read it and I understand how the image was created. I was concerned the image was created when modifying the new iframe CSS file, because that would be weird.

Haven't tested but the code/approach looks the best.

(I'm looking forward to not being sick and having intelligence again...)

Issue #878, opened 2019-02-20
Pixel Viewer Interactive displaying blank white image


eAlasdair:
Browser: Firefox
URL: https://cs-field-guide-dev.appspot.com/en/interactives/pixel-viewer/

When loading the Pixel Viewer Interactive; the image is zoomed in on, then the menu fades in and simultaneously _the image fades to white_

Subsequently no interactions with the interactive beyond opening/closing the menu result in a visible change

![v](https://user-images.githubusercontent.com/25916475/53129475-8277ca80-35cc-11e9-8fd8-5a20fdbb9923.png)

This and issue #880 could be resolved together

JackMorganNZ:
Error log:

```
Uncaught DOMException: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': The canvas has been tainted by cross-origin data.
    at get_pixel_data (https://storage.googleapis.com/cs-field-guide-dev.appspot.com/static/interactives/pixel-viewer/js/pixel-viewer.js:846:48)
    at paint (https://storage.googleapis.com/cs-field-guide-dev.appspot.com/static/interactives/pixel-viewer/js/pixel-viewer.js:946:19)
    at Tiling.render (https://storage.googleapis.com/cs-field-guide-dev.appspot.com/static/interactives/pixel-viewer/js/third-party/Tiling.js:110:4)
    at Scroller.render [as __callback] (https://storage.googleapis.com/cs-field-guide-dev.appspot.com/static/interactives/pixel-viewer/js/pixel-viewer.js:912:12)
    at Scroller.__publish (https://storage.googleapis.com/cs-field-guide-dev.appspot.com/static/interactives/pixel-viewer/js/third-party/Scroller.js:1123:11)
    at Scroller.scrollTo (https://storage.googleapis.com/cs-field-guide-dev.appspot.com/static/interactives/pixel-viewer/js/third-party/Scroller.js:599:9)
    at Image.image.onload (https://storage.googleapis.com/cs-field-guide-dev.appspot.com/static/interactives/pixel-viewer/js/pixel-viewer.js:869:18)
pixel-viewer.js:846 Uncaught DOMException: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': The canvas has been tainted by cross-origin data.
    at get_pixel_data (https://storage.googleapis.com/cs-field-guide-dev.appspot.com/static/interactives/pixel-viewer/js/pixel-viewer.js:846:48)
[...]

PR #956, opened 2019-05-16
Prevent canvas from being tainted by setting the crossorigin attribute in the JavaScript..

courtneycb:
In #955 we set the crossorigin attribute to 'anonymous' in the HTML templates. However we also needed to set this on canvas elements in the JavaScipt files (for the canvas elements that get canvas methods run on them - such as getImageData).

Will not know if this resolves our problem until it is merged into develop..

Relates to #878

### DIFF
--- a/csfieldguide/static/interactives/image-bit-comparer/js/image-bit-comparer.js
+++ b/csfieldguide/static/interactives/image-bit-comparer/js/image-bit-comparer.js
@@ -259,6 +259,7 @@ function loadImage() {
 function initialCanvasData() {
   var source_canvas = document.getElementById('interactive-image-bit-comparer-source-canvas');
   var source_canvas_context = source_canvas.getContext('2d');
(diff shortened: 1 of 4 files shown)


csfieldguide/static/interactives/image-bit-comparer/js/image-bit-comparer.js
```diff
@@ -259,6 +259,7 @@ function loadImage() {
 function initialCanvasData() {
   var source_canvas = document.getElementById('interactive-image-bit-comparer-source-canvas');
   var source_canvas_context = source_canvas.getContext('2d');
+  source_canvas_context.crosOrigin = 'anonymous';
```
eAlasdair: spelling mistake? `crosOrigin`

courtneycb: nice spotting

PR #957, opened 2019-05-20
At the end of each chapter, link to the next

### LINKED ISSUE #884: At the end of each chapter, link to the next
Now that there is no longer a handy navigation tab to one side of the page, we should add links to the following chapter somewhere on each chapter. On the following example, such a link could be positioned opposite `Previous: about this guide` on the bottom right of the main window on the final page of the chapter

v

eAlasdair:
Resolves #884 

image

## Limitations

For any chapter number `x`, the end of the chapter will attempt to link to chapter number `x+1`. If it doesn't find one it doesn't make the link, even if for example chapter number `x+2` does exist. On the current version of the dev site (at time of writing), chapters 3 (prog lang) and 13 (comp vision) are commented out as they are not ready for release. As such chapters 2 and 12 (as well as the final one of course) won't have links to the next chapters. This is only a minor issue because we can expect that all chapters will be available on release.

If there is a missing chapter section number an error is thrown during the build/update process, so it will never link to the next chapter with chapter sections still remaining

image

## Also

While doing this I noticed several sentences without `trans` tags, didn't add them in case they are done elsewhere somehow.

### DIFF
--- a/csfieldguide/chapters/views.py
+++ b/csfieldguide/chapters/views.py
@@ -63,7 +63,7 @@ class ChapterSectionView(generic.DetailView):
             ChapterSection object (ChapterSection).
 
         Raises:
-           404 error: if oject cannot be found.
+           404 error: if object cannot be found.
         """
         return get_object_or_404(
             self.model.objects.select_related(),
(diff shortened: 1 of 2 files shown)


eAlasdair:
> Nice work. Did you make note of where the missing 'trans' tags were?

`chapters/*`
`general/index.html`
among others

PR #959, opened 2019-05-22
Add django translation tags to base html files

eAlasdair:
This brings all .html files in the `templates` folder in line with our intention to have all pages translatable, excluding individual interactives which are dealt with elsewhere

RESOLVED
How do I deal with this line in `search/search.html`?
```<p class="text-center"><em>Showing {% if is_paginated %}results {{ page_obj.start_index }} to {{ page_obj.end_index }} of {% endif %}{{ paginator.count }} results</em></p>```

Also:
Note that in page_heading blocks sometimes the text has `h1` tags, and sometimes it doesn't

### CODE REVIEW

File: csfieldguide/templates/chapters/chapter_section.html [hunk 1]
```diff
@@ -29,16 +29,16 @@
   <div>
     {% if previous_section %}
       <a href="{% url 'chapters:chapter_section' chapter.slug previous_section.0.slug %}" class="btn btn-primary float-left mb-2 {% if not chapter_section.translation_available %} text-muted{% endif %}">
-        Previous: {{ previous_section.0.name }}
+        {% trans 'Previous:' %} {{ previous_section.0.name }}
```
courtneycb: I think we also need to include `{{ previous_section.0.name }}` inside the trans tag if it allows you? I think the translators will understand what the variable means. The order may be different in different languages. This could also relate to your 'help-wanted' question?

If it breaks when you wrap it then ignore me :smiley_cat:

csfieldguide/templates/general/index.html
```diff
@@ -17,15 +17,15 @@
...
           </h1>
           <p class="lead">
-            An online interactive resource for high school students learning about computer science.
+            {% trans 'An online interactive resource for high school students learning about computer science.' %}
           </p>
           <p>
             <small>
               <a class="text-white" href="http://cs-field-guide.readthedocs.io/en/latest/changelog.html">
-                Currently {{ VERSION_NUMBER }}
+                {{ VERSION_NUMBER }}
```
courtneycb: If the wrapping thing I talked about before works you could probably add back in 'Currently' here

csfieldguide/templates/chapters/chapter_section.html
eAlasdair: It does break. I don't think it's an issue because it's next-*colon*-chaptername, so the order shouldn't change

csfieldguide/templates/general/index.html
eAlasdair: Yeah it breaks, prints `Currently {{ VERSION_NUMBER }}` instead of the actual number

csfieldguide/templates/chapters/chapter_section.html
courtneycb: Does this link help at all? https://stackoverflow.com/questions/6546933/django-translate-variable-content-in-template

eAlasdair: It does not

courtneycb: :crying_cat_face:

courtneycb: We may need to summon @JackMorganNZ

csfieldguide/templates/general/index.html
courtneycb: Lets remove this as it is displayed in the footer

courtneycb:
Spoke with Jack, Search doesn't work for any language other than English so we don't need to translate that line you want help with.

PR #977, opened 2019-05-29
Open links to interactives in a new tab

### LINKED ISSUE #974: Investigate opening interactive in new tab or providing back link
_**This issue is user submitted and needs to be validated:**_

I love that this resource goes from strength to strength and continues to be updated.  One thing that I noticed was that when I clicked on an interactive in the 'Algorithms' chapter, it took me to the working interactive but there was no easy way to get back to the chapter.  When I used my browser's 'back' button it took me to the start of the chapter which is not ideal.

I'm wondering if it would be possible to have interactions open in a new tab by default as that would solve the problem.

courtneycb:
This change will cause links to interactives to open in a new tab ~~(unless overridden - see below).~~

This behaviour is being introduced because users will often want to refer to the chapter text while they are using the interactive. Also currently if the user hits the back button once in an interactive, they will be taken to the start of the chapter section they were on, not to the bit they were up to when they clicked the interactive link which can be irritating.

~~However we do not want this behaviour if the user is accessing the interactive from, for example, the `appendices/interactives` page. Thus we have introduced the argument `new_tab` for the `render_interactive_link` template tag. By default this argument is `True` for all interactive links but can be set to `False` when we do not want a new tab to be opened. Instead it will use the default behaviour of opening the interactive in the same frame it was clicked.~~

Resolves #974


eAlasdair:
Kinda related to #783

eAlasdair:
So the default behaviour is to overwrite the default behaviour? Okay I understand

courtneycb:
> I think we should be consistent and open all interactives in a new page, even from the interactives page

This was a suggestion from @JackMorganNZ so you guys might want to have a discussion on what we should do. I don't have any strong feelings for either, although the user might expect it to open in a new tab since the other links do.

Issue #454, opened 2017-06-14
Some images are too large


hayleyavw:
The pictures are currently HUGE

eAlasdair:
custom html has already been written `csfieldguide/utils/custom_converter_templates/image.html`

eAlasdair:
cusom css has already been written as `.img-container` in `csfieldguide/static/scss/website.scss` but the images are still huge under certain circumnsances

eAlasdair:
- Images scale to fill most of the width of the screen, this is not inherently a bad thing because it suits some wide but not tall images such as the substitution cipher ones [here](https://cs-field-guide-dev.appspot.com/en/chapters/coding-encryption/substitution-ciphers/)
- However, on desktop screens, images (and videos) above a certain height practically fill the screen, such as the cartoons [here](https://cs-field-guide-dev.appspot.com/en/chapters/software-engineering/)
- We need a conditional, so that only images above a certain height are scaled to a more appropriate max width
- As far as I know, that can't be done in css
- The image template `csfieldguide/utils/custom_converter_templates/image.html` is an html file but is interpreted differently, so as far as I could find it doesn't allow JS of any form

JackMorganNZ:
This was a topic we spent a lot of time on deciding how to solve for CS Unplugged, I can share our approaches when I return to work.

> * As far as I know, that can't be done in css

If I understand what you are asking for correctly, this can be done with CSS media queries.

eAlasdair:
Current solution is to scale down the source images

JackMorganNZ:
This task involves two steps:
- [ ] Checking image size (resolution)
- [ ] Checking image size (file size)

eAlasdair:
None of the images are in the MBs (like #512)

Excluding ones associated with individual interactives, I checked all the images and only the top 4 largest are above 200kb:
- `compression-comparison.png` (540kb)
- `compressed-jpeg.png` (ironically) (441kb)
- `fruit-bowl-photo-with-canny-edge-detection.png` (215kb)
- `time-capsule.png` (206kb)

eAlasdair:
Closing this issue as I believe it has been addressed. Reopen if further action is needed

PR #982, opened 2019-05-31
scale down images

eAlasdair:
Relates to #454 

Scales several images down to smaller dimensions to fit more nicely in the content. In many cases this was to reduce the width to 500px and scale the height appropriately


courtneycb:
Really awesome work! Makes everything look *much* neater. I think these two comics below look a little fuzzy, is it like that for you? 

image

eAlasdair:
Images are enhanced and, on average, a smaller file size, so that's good

PR #986, opened 2019-05-31
Remove broken link

eAlasdair:
This link in the NCP chapter is a 404 error, but Linkie did not find it
https://learn.code.org/s/1/level/102

*Issue brought up by a user 20/10/2018*


eAlasdair:
> Perhaps check with Tim if this could be a replacement link:
> 
> https://code.org/curriculum/course3/18/Teacher

Tim approved that link, has been added as replacement

PR #1001, opened 2019-06-09
1001 ways to Recreate dot-combinations as interactive

### LINKED ISSUE #219: Braille interactive for data representation
Comment removed from chapter text

there's room for some interactives with the braille, something like getting students to try to find all the patterns for 3, 4, 5, 6 dots (as a bit of a game) [low priority]

eAlasdair:
Resolves #219 by recreating the dot-combinations uninteractive as an interactive activity

### DIFF
--- a/csfieldguide/templates/interactives/dot-combinations.html
+++ b/csfieldguide/templates/interactives/dot-combinations.html
@@ -136,3 +136,7 @@
 {% block css %}
   <link rel="stylesheet" href="{% static 'interactives/dot-combinations/css/dot-combinations.css' %}">
 {% endblock css %}
+
+{% block js %}
+  <script type="text/javascript" src="{% static 'interactives/dot-combinations/js/dot-combinations.js' %}"></script>
+{% endblock js %}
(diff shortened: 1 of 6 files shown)


csfieldguide/templates/interactives/dot-combinations.html
```diff
@@ -4,135 +4,59 @@
...
+    <!-- Heading and Description -->
+    <div class="col-12">
+      <h3>{% trans "Dot Combinations" %}</h3>
+    </div>
+    <div class="col-12 mb-3">
+      {% trans 'Click on a dot to flip it. Try to find all the combinations.' %}
+    </div>
+    <div class="form-group col-12 col-sm-4 offset-sm-4">
+      <label for="number-of-dots">{% trans 'Number of dots:' %}</label>
+      <select class="form-control-md" id="number-of-dots" value="3">
```
courtneycb: Really strange but only form-control-lg and form-control-sm exist, I think `form-control` is just medium by default.

eAlasdair: I don't know how but `form-control-md` changed the input box from a standard input dropdown to something I really liked (image) so i want to keep it that way. Does it show like this on chrome?
image

courtneycb: Unfortunately it shows like this:

image

PR #1005, opened 2019-06-10
add configuration options to pixel viewer

### LINKED ISSUE #964: Let the user pick the mode for pixel viewer if it's loaded without url parameters.
Suggestion from @eAlasdair 

> Also could you see if you can add functionality to let the user pick the mode they use if it's loaded without url parameters? So like 2d-shape-manipulations users are held to a specific mode if the interactive is opened in the chapter, but if they open it from the 'interactives' page they get a dropdown or something so they can load the other modes

eAlasdair:
Resolves #964, though I didn't worry about having the option only appear if the interactive was loaded without a config

### DIFF
--- a/csfieldguide/templates/interactives/pixel-viewer.html
+++ b/csfieldguide/templates/interactives/pixel-viewer.html
@@ -22,10 +22,10 @@
             <canvas id="pixel-viewer-interactive-content" crossorigin="anonymous"></canvas>
             <div id="pixel-viewer-interactive-settings" class="menu-offscreen p-3">
                 <h3 id="pixel-viewer-interactive-title">
-                    {% trans "Pixel Value Interactive" %}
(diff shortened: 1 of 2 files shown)


csfieldguide/templates/interactives/pixel-viewer.html
```diff
@@ -22,10 +22,10 @@
             <canvas id="pixel-viewer-interactive-content" crossorigin="anonymous"></canvas>
             <div id="pixel-viewer-interactive-settings" class="menu-offscreen p-3">
                 <h3 id="pixel-viewer-interactive-title">
-                    {% trans "Pixel Value Interactive" %}
+                    {% trans "Pixel Viewer Interactive" %}
```
courtneycb: Lol pixel value interactive??? Nice spotting!!

PR #1006, opened 2019-06-10
Align chapter section buttons better if they don't fit in one line

### LINKED ISSUE #949: Chapter section buttons don't align nicely when the title is long.
Need suggestions on what we should do here.

image

image

eAlasdair:
Resolves #949 
The d-flex options override justify-left/-right when applicable. This allows the buttons to maintain the same height if one of them has to wrap text

### DIFF
--- a/csfieldguide/static/scss/website.scss
+++ b/csfieldguide/static/scss/website.scss
@@ -95,6 +95,15 @@ body {
   cursor: pointer;
 }
 
+.btn-chapter-section {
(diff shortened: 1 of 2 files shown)


eAlasdair:
image

eAlasdair:
> Approving as is. If @JackMorganNZ would like further changes we will add them in a seperate PR as per our discussion on slack @eAlasdair :)

By as is we mean after a chage

PR #1008, opened 2019-06-11
Remove docker volumes when the ./csfg end command is run

### LINKED ISSUE #906: Docker Volumes are never removed
Every time the command `./csfg start` is run, Docker creates two new Volumes - one for Django, and one for Nginx. These should probably be removed by `./csfg end`, but are not, nor are they removed by `./csfg clean`.

We should implement functionality that removes unused Volumes

In the interim, if the multitudes of Volumes are affecting available storage space, this command will remove **ALL** Docker volumes (*including those unrelated to the Field Guide*):

`docker volume rm $(docker volume ls -q)`

And this will remove as much of everything (else) as possible:
`docker system prune -a`

courtneycb:
Resolves #906

### DIFF
--- a/csfg
+++ b/csfg
@@ -60,6 +60,12 @@ defhelp start 'Start development environment.'
 cmd_end() {
   echo "Stopping systems... (takes roughly 10 to 20 seconds)"
   docker-compose down
+  echo
+  echo "Deleting system volumes..."
+  volumes=($(docker volume ls -qf dangling=true ))
+  for volume in "${volumes[@]}"; do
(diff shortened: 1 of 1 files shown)


eAlasdair:
Something to check though is that it only removes csfg volumes, I assume we don't want to remove csu stuff for example

courtneycb:
> Something to check though is that it only removes csfg volumes, I assume we don't want to remove csu stuff for example

Have checked and it only removes csfg volumes.

PR #1012, opened 2019-06-14
Format heading numbers

JackMorganNZ:
This pull request formats the heading numbers to be less obtrusive. It also adds a permalink to the title next.

**Example:**
localhost_81_en_chapters_computer-graphics_drawing-lines-and-circles_

Other notes:

- The permalink is created using the heading text.
- A checker has been added to ensure two heading on the same page don't have the same permalinks.
- The permalink could have been created using the number, however this could lead to permalinks linking to different content when new sections are added beforehand.
- The permalink could also have been created using both the title and numbers, however the numbers are more likely to change, therefore only the text is used.
- Headings with numbers longer than 3 levels now don't show a number.
- When hovering over a title, a permalink icon is displayed next to it.


csfieldguide/utils/check_heading_tree.py
```diff
@@ -0,0 +1,20 @@
+"""Check for any duplicate headings in the heading tree."""
+
+from utils.errors.DuplicateHeadingFoundInMarkdownFileError import DuplicateHeadingFoundInMarkdownFileError
+
+
+def check_heading_tree(heading_tree, md_file_path):
```
eAlasdair: Travis is failing because this and traverse_node are missing docstrings

JackMorganNZ: Thanks, this is now fixed.

csfieldguide/utils/custom_converter_templates/chapter-section/heading.html
```diff
@@ -1,22 +1,20 @@
-<{{ heading_type }} id="{{ title_slug }}">
-<span class="section_number">
+<div class="section-heading">
+{%- if heading_level < 4 -%}
+<div class="section-heading-number">
```
eAlasdair: >     * Headings with numbers longer than 3 levels now don't show a number.

I don't think this is taking into account the chapter number itself, so if `Headings with numbers longer than 3 levels now don't show a number` it would only be # and ## headings in markdown that get numbers. Hope that makes some sense, point is; *headings numbers as long as 4 levels of numbers still show a number*

image

JackMorganNZ: Nice catch, will fix it later today.

eAlasdair: You seem very busy @JackMorganNZ so I fixed it. Should be good now

csfieldguide/utils/errors/DuplicateHeadingFoundInMarkdownFileError.py
```diff
@@ -0,0 +1,26 @@
...
+        super().__init__()
+        self.md_file_path = md_file_path
+
+    def __str__(self):
+        """Override default error string.
+
+        Returns:
+            Error message for missing heading.
+        """
+        return self.base_message.format(filename=self.md_file_path) + ERROR_MESSAGE
```
eAlasdair: I think this is the line not hit according to codecov

PR #1022, opened 2019-06-24
Revert syntax for translated Verto tags in Spanish

### LINKED ISSUE #969: Markdown comments are being translated
image

We need to ensure Translators understand not to translate `{comment}` tags, or remove them entirely

JackMorganNZ:
Comments that should not be shown are being shown to users due to a Verto tag being translated. In future all comment strings needs to be disabled on Crowdin.

Fixes #969


JackMorganNZ:
Yep, all strings starting with `{comment` have been hidden.

PR #1025, opened 2019-06-24
Show thumbnails for all languages

### LINKED ISSUE #905: Interactive thumbnails don't load in non-English languages
Thumbnails attempt to load from `.../thumbnails/[language-code]/[interactive-name]`
However thumbnails are only generated for `.../thumbnails/en/[interactive-name]`

We need to either generate thumbnails for all languages, or request the English version in all languages

There's a chance that this is just because the interactives haven't been translated yet

JackMorganNZ:
It also only shows thumbnails for interactives available in the viewed language.

Fixes #905.

Branch has wrong number but that doesn't really matter.


eAlasdair:
Thumbnails take a significant amount of time to be created, and we plan on having this guide in many languages, so is there a way to only have them all generated in the production build?

JackMorganNZ:
We could have it only load English for local update, and then all languages for deployment. But it's probably best for it to load all on `update`. The system will take longer and longer to load as more languages are added, and we should just run the sub-commands we need when required, and update is used for the full website load. Thoughts?

eAlasdair:
> We could have it only load English for local update, and then all languages for deployment. But it's probably best for it to load all on `update`. The system will take longer and longer to load as more languages are added, and we should just run the sub-commands we need when required, and update is used for the full website load. Thoughts?

I agree it should be everything for update, but in that case I strongly suggest creating a new semi-mainstream command (like update-static) that does not do it. During development the thumbnails almost never need to be regenerated

JackMorganNZ:
> In fact I suggest that the command **only** reloads the content, if possible (like how update-static only updates the staticfiles). That would make checking changes to markdown files faster

Which command are you referring to here?

Also `updatedata` exists for just updating the Markdown, plus we can add an `update_lite` command like CS Unplugged has.

JackMorganNZ:
Also we can resolve which commands trigger this after the 3.0 release.

eAlasdair:
> > In fact I suggest that the command **only** reloads the content, if possible (like how update-static only updates the staticfiles). That would make checking changes to markdown files faster
> 
> Which command are you referring to here?
> 
> Also `updatedata` exists for just updating the Markdown, plus we can add an `update_lite` command like CS Unplugged has.

I was referring to this new command I'm suggesting, but if updatedata does that then it's not needed. Though for consistency shouldn't it be update_data?

> Also we can resolve which commands trigger this after the 3.0 release.

Yes, that's fine

PR #1028, opened 2019-06-25
Redirect old links for interactives to 3.0 interactives

### LINKED ISSUE #1027: Redirect old links for interactives to 3.0 interactives
#872 redirects all old URLs to 3.0 content **except** for interactives.

We should add redirects for these.

courtneycb:
This PR redirects links for interactives in the 2.0 site to interactives in the 3.0 site.

Resolves #1027


eAlasdair:
Doesn't work for mips interactives because the old ones are php, is that an issue?

PR #1032, opened 2019-06-26
Chapter content and navigation update

JackMorganNZ:
## Content improvements

- Splits most chapters introduction into a short introduction and a "What's the big picture?" section.
  - This was done to keep the introduction short, and prevent it from containing headings that would be numbered in an odd fashion (`3.0.1`).
    - The template for numbering of chapter headings is removed in case there is headings.
  - This split was done before the glossary entries were added, and I had to merge that PR into this work, I believe it's done correctly, but worth checking.

These changes were done to firstly improve the readability of content, but also to allow the following:

## Navigation improvements

- Chapter section headings are now stored in the database.
  -  The loader iterates through the headings for each language, and for each heading/language pair, it creates an heading object in the database. This function uses a different approach compared to the standard `TranslatableModel` and `TranslatableModelLoader` approach, for the following reasons:
      - Heading order for a section can be different across languages due to translations after initial release of section. This may not be a problem due to numbering, but I'm still unsure and would require thorough testing.
      - Slugs are different across languages (due to Verto processing files individually.
      - Source of the data coming from VertoResult objects, instead of separate files.

## Misc

- Removes JavaScript for moving page to avoid anchor links being covered by fixed navbar, since content element is after the navbar (structured by flexbox). This JavaScript is not required.
  - The way the content is structured seems to not allow all common ScrollSpy libraries to function (Bootstrap ScrollSpy and Gumshoe). It may be worth considering implementing our own method using `Element.getBoundingClientRect()` or Intersection Observer API, or reverting our layout to allow established libraries.
- Fixes a few docstrings.
- Enables native CSS smooth scrolling between anchors for new browsers that support it, and don't have `prefers-reduced-motion` flag.
- Labels languages other than English as `Beta`, due to partial translation and support for 3.0 release.
- Center chapter video above content.
- Adds chapter section navigation near top of chapter page.


csfieldguide/templates/base.html
```diff
@@ -60,6 +60,9 @@
                 {% get_language_info for language.0 as lang %}
                   <a class="dropdown-item{% if LANGUAGE_CODE == lang.code %} active{% endif %}" href="{% translate_url language.0 %}">
                     {{ lang.name_local|capfirst }}
+                    {% if lang.code != "en" %}
+                      <span class="badge badge-warning">Beta</span>
```
eAlasdair: Beta should be translated - for the languages that don't use greek

csfieldguide/chapters/content/en/algorithms/sections/whats-the-big-picture.md
```diff
@@ -0,0 +1,148 @@
...
+That's worth knowing in advance because we usually need our programs to scale up well; in the case of the high scores, if you're running a game that suddenly becomes popular, you want to know in advance that the high score algorithm will be fast enough if you get more scores to check.
+
+{panel type="extra-for-experts"}
+
+# Algorithm complexity
+
+The formal term for working out the cost of an algorithm is [algorithm analysis](https://en.wikipedia.org/wiki/Analysis_of_algorithms), and we often refer to the cost as the algorithm's *complexity*.
+The most common complexity is the "time complexity" (a rough idea of how long it takes to run), but often the "space complexity" is of interest - how much memory or disk space will the algorithm use up when it's running?
+
+There's more about how the cost of an algorithm is described in industry, using a widely agreed on convention called 'Big-O Notation', in the ["The whole story!"](the-whole-story) section at the end of this chapter.
```
eAlasdair: These changes seem to be putting even more restriction on how links are interpreted. This link breaks now so the page doesn't load

csfieldguide/templates/chapters/chapter_section.html
```diff
@@ -46,14 +46,21 @@
...
         {% for section in chapter.chapter_sections.all %}
           <li class="nav-item">
             <a class="nav-link {% if section.slug == chapter_section.slug %} active text-white{% else %} nav-link-colour{% endif %}" href="{% url 'chapters:chapter_section' chapter.slug section.slug %}">
               {{ section.name }}
             </a>
+            {% if section.slug == chapter_section.slug %}
+              <nav id="section-headings" class="nav nav-pills flex-column my-1 ml-3">
+                {% for heading in headings %}
+                  <a class="nav-link" href="#{{ heading.slug }}">{{ heading.name }}</a>
+                {% endfor %}
```
eAlasdair: I'm guessing it's one of the scrollspy things that couldn't be implemented, but I'll point out anyway that it's jarring to click a heading in the table of contents, have the page jump to that heading, and not have the heading I clicked on highlighted

csfieldguide/chapters/content/en/algorithms/sections/whats-the-big-picture.md
courtneycb: The syntax is just incorrect.

JackMorganNZ: Sorry, old code that snuck through the merge conflict.

csfieldguide/templates/chapters/chapter_section.html
JackMorganNZ: Yes, Scrollspy does all that. Going to look into it later.

courtneycb:
Just some suggestions:

-  We should remove the 'sections in this chapter' part as it duplicates information. If we do remove this we might want to put the chapter section numbers on the sidebar nav.

image

- We should link back to the main chapter page. E.g for the sidebar below we would include a link to 'Data Representation'. We could replace 'Table of contents' with this link.

image

csfieldguide/templates/chapters/chapter_section.html
JackMorganNZ: Now fixed

courtneycb:
I get this strange behaviour where the the nav bar will highlight the subsection below the one we are currently viewing. Screen recording below.

out-1

eAlasdair:
Both issues @courtneycb mentioned are happening for me also. For the former issue, what's happening is the final subsection highlights if it is visible on the page, regardless of there being other subsections visible

JackMorganNZ:
> Both issues @courtneycb mentioned are happening for me also. For the former issue, what's happening is the final subsection highlights if it is visible on the page, regardless of there being other subsections visible

I've discovered that this is an intended feature: https://github.com/cferdinandi/gumshoe/issues/56#issuecomment-273156295

Possible solutions:
- Lower the point where the current heading is selected. I'm using 300px at the moment, could make it half the viewport height.
- Add larger spacing above headings.

PR #1038, opened 2019-06-27
Reformat link-items to nav-pill style

JackMorganNZ:
This will match the style used in PR #1032.

Also stacks chapter links vertically for easier readability, and removes the 'Coming soon' text (as it's vague and could be misleading, chapters come when they are ready).


eAlasdair:
Personally I really dislike these changes, as there is now a ton of white space everywhere. Could we have the interactive thumbnails larger on the index page, and the chapters take two columns if the screen is big enough?

PR #1039, opened 2019-06-27
Add links to Polish version of 2.6.0

eAlasdair:
Adds links to polish standard ~~and teacher~~ versions of the 2.6.0 Field Guide


eAlasdair:
Gah I used the wrong links

csfieldguide/templates/base.html
```diff
@@ -62,6 +62,12 @@
                     {{ lang.name_local|capfirst }}
                   </a>
                 {% endfor %}
+                <a class="dropdown-item" href="https://bezkomputera.wmi.amu.edu.pl/ppi/">
+                  język polski (2.6.0)
+                </a>
+                <a class="dropdown-item" href="https://bezkomputera.wmi.amu.edu.pl/ppi/teacher/">
```
courtneycb: Jack just told me we will send them an email telling them to add a link to the teacher version on their homepage. So we don't need to link to the teacher version.

Can we also apply the `text-muted` css class to the version number :)

PR #1058, opened 2019-07-01
URL redirects

courtneycb:
This PR redirects https://csfieldguide.org.nz/index.html to https://csfieldguide.org.nz/ and tidies some code in the `urls.py` files.

~~HELP WANTED:~~

- Currently the redirects for permalinks are incorrect. For example `/chapters/algorithms.html#searching` should redirect to `/chapters/algorithms/searching/` but instead it redirects to `http://localhost:81/en/chapters/algorithms/#searching`.

- Here are the two redirects I found that do not work:
In `chapters/urls.py`:
```
    # eg: redirect /chapters/algorithms.html#searching to /chapters/algorithms/searching/
    url(
        r"^(?P<chapter_slug>[-\w]+).html#(?P<chapter_section_slug>[-\w]+)$",
        RedirectView.as_view(permanent=True, pattern_name="chapters:chapter_section"),
    ),
```
In `curriculum_guides/urls.py`:
```
    # eg: redirect /curriculum-guides/apcsp/index.html#abstraction to /curriculum-guides/apcsp/abstraction/
    url(
        r"^(?P<curriculum_guide_slug>[-\w]+)/index.html#(?P<curriculum_guide_section_slug>[-\w]+)$",
        RedirectView.as_view(permanent=True,
                             pattern_name="curriculum_guides:curriculum_guide_section"),
    ),
```

I tried overriding the `get_url_redirect` method in `views.py` but the code would never get hit? Not sure what's going on and don't seem to be getting very far with it.

### DIFF
--- a/csfieldguide/chapters/urls.py
+++ b/csfieldguide/chapters/urls.py
@@ -40,7 +40,7 @@ urlpatterns = [
     # eg: redirect /chapters/index.html to /chapters/
     url(
         r"^index.html$",
-        RedirectView.as_view(permanent=True, url="/chapters/"),
+        RedirectView.as_view(permanent=True, pattern_name="chapters:index"),
     ),
     # eg: redirect /chapters/algorithms.html to /chapters/algorithms/
     url(
@@ -50,6 +50,6 @@ urlpatterns = [
     # eg: redirect /chapters/algorithms.html#searching to /chapters/algorithms/searching/
     url(
         r"^(?P<chapter_slug>[-\w]+).html#(?P<chapter_section_slug>[-\w]+)$",
-        RedirectView.as_view(permanent=True, pattern_name="chapters:chapter:chapter_section"),
+        RedirectView.as_view(permanent=True, pattern_name="chapters:chapter_section"),
     ),
 ]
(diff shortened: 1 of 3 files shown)


eAlasdair:
Did some research. Anchor parameters (the bit after the #) do not get sent as part of the request, so the server never knows about them. Your code never got hit because those redirects never got hit. There is no way those redirects can do what they are attempting, so they can be removed.

courtneycb:
> Did some research. Anchor parameters (the bit after the #) do not get sent as part of the request, so the server never knows about them. Your code never got hit because those redirects never got hit. There is no way those redirects can do what they are attempting, so they can be removed.

That's a pain. Thanks for doing that research. Should we re-word sentences so the user doesn't expect to be taken directly to the section?

eAlasdair:
> > Did some research. Anchor parameters (the bit after the #) do not get sent as part of the request, so the server never knows about them. Your code never got hit because those redirects never got hit. There is no way those redirects can do what they are attempting, so they can be removed.
> 
> That's a pain. Thanks for doing that research. Should we re-word sentences so the user doesn't expect to be taken directly to the section?

Which sentences? Any current content will take the user directly to the section, because it uses the current style. It's just links to the old version can only link to the introduction page of the new version

courtneycb:
> Which sentences? Any current content will take the user directly to the section, because it uses the current style. It's just links to the old version can only link to the introduction page of the new version

Oh man, ignore me. Confused about something else.

PR #1079, opened 2019-07-04
Align and resize homepage logos.

### LINKED ISSUE #1050: Homepage images not aligned correctly
The project icon and InternetNZ images are not aligned or sized correctly. This is more obvious on a mobile device.

courtneycb:
Resolves #1050


courtneycb:
> This solution to this problem for _all_ viewports is to make the SVG scale responsively, using a viewbox attribute.

These are PNGs?

JackMorganNZ:
> > This solution to this problem for _all_ viewports is to make the SVG scale responsively, using a viewbox attribute.
> 
> These are PNGs? Nvm found the SVG for the CSFG logo.

Sorry, I thought the homepage logo was loaded as an SVG but it appears to be a PNG. But should still be resized responsively.

eAlasdair:
Logo now gets resized responsively, though is now generally smaller on smaller screens because it no longer overflows the boundary of its container

csfieldguide/templates/general/index.html
```diff
@@ -10,10 +10,10 @@
   <div class="jumbotron jumbotron-fluid py-5">
     <div class="container text-center">
       <div class="row align-items-center">
-        <div class="col-6 mx-auto col-md-4 order-md-2 csfg-logo-center">
-          <img id="homepage-logo" src="{% static 'img/logos/guide-logo-light.png' %}">
+        <div class="col-6 mx-auto col-md-4 order-md-2">
+          <img class="csfg-logo-image" id="homepage-logo" src="{% static 'img/logos/guide-logo-light.png' %}">
```
JackMorganNZ: This seems to be an odd way to select the image for CSS. Why use a class (which is used for multiple elements), when it already has an ID (an unique identifier)?

eAlasdair: It was a class originally so I stuck with that. Will change

eAlasdair: Misclicked mid-commit, but have added the attributes to _homepage.scss

csfieldguide/templates/base.html
```diff
@@ -137,7 +137,7 @@ <h1>
                     <img class="img-fluid py-5" src="{% static 'img/logos/uc-computer-science-education-logo.png' %}">
                   </a>
                 </div>
-                <div class="col-12 col-sm-6 col-md-4 text-center">
+                <div class="col-12 col-sm-6 col-md-4 logo-mobile-center">
```
JackMorganNZ: This seems to do nothing new. Can you please explain it's purpose?

courtneycb: I couldn't find a bootstrap class that would apply the `text-center` class for viewports size SM or _smaller_. I could only find `text-sm-center` which will apply for viewports of size SM or _larger_. Do you know how to achieve this with built in bootstrap classes?

JackMorganNZ: Bootstrap has appeared (can't find logic for decision) to only do viewports going up. Either an earlier version or another library I used (Bootstrap 2 or MaterializeCSS) had both directions but this can be confusing.

Regardless, they way to do it is `text-center text-md-left`. Theoretically it would mean we don't need to include the new rule and save a few bytes in the CSS file, but up to you if you want to change it.

courtneycb: Change made! :)

PR #1087, opened 2019-07-09
Add interactive to demonstrate limitations of short term memory

courtneycb:
This PR introduces a new interactive to demonstrate the limitations of short term memory.

**Notes:**

- ~~Users could enter alternative words for items that would technically be correct such as 'bicycle' for 'bike' , however the way the interactive is currently implemented it will not recognise these words as correct.~~

- I added wording to introduce the interactive but am open to suggestions as English definitely isn't my strong point.

TODO:
- [x] Investigate adding more images and randomly picking some to show.
- [x] Add link to interactive in content.
          - **Need help with wording here on introducing interactive.**
- [x] Rename interactive

Resolves #144

PR #1090, opened 2019-07-12
Add second short term memory interactive

eAlasdair:
During the development of #1087 I had another idea for a way to demonstrate the limitations of STM.
Number Memory focuses on remembering one number through distractions, rather than memorizing images in a time period. It is designed to work in-page and has been added in context
image


csfieldguide/static/interactives/number-memory/js/number-memory.js
```diff
@@ -0,0 +1,202 @@
+const STALL_TEXT = [
+  [1000, gettext("Finding some paper...")],
+  [300, gettext("*BONK*")],
+  [500, gettext("Ow!")],
+  [1000, gettext("Bumped my head.")],
+  [1500, gettext("Got it, now where was that pen?")],
+  [300, gettext("Ah ha!")],
+  [500, gettext("*Tap tap tap*")],
```
courtneycb: Maybe this could be `*scribble scribble*` as well? I think it makes more sense to do the same action (as you do later on) to check if the pen is working :) Also maybe expand on `Empty. :(`, maybe even just `It's empty :(`.

csfieldguide/chapters/content/en/human-computer-interaction/sections/interface-usability.md
```diff
@@ -109,6 +109,8 @@ If a system takes some time to respond (say, 10 seconds) then chances are the us
 For example, if you have a phone number to type in that someone has just told you, and it takes 12 seconds before you can type it, you may forget the number, whereas if you can access the interface within a couple of seconds, you can probably enter the number without much effort.
 For this reason, any part of a system that takes more than about 10 seconds to respond forces the user to rehearse or write down key information, which is more tiring.
 
+{interactive slug="number-memory" type="in-page"}
```
courtneycb: Maybe add some sort of text to introduce the interactive (don't have any suggestions right now).

eAlasdair: The text immediately above is perfect for it, so I don't need to add anything

csfieldguide/static/interactives/number-memory/js/number-memory.js
```diff
@@ -0,0 +1,202 @@
...
+  }
+}
+
+/**
+ * Hides irrrelevant HTML divs, then reveals relevant ones.
+ * Also resets the number input box and sets the title to the remember text.
+ */
+function showStartScreen() {
+  $('#number-input').val('');
+  $('#number-memory-title').html(TXT_REMEMBER);
```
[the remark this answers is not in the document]
eAlasdair: I think that's self-explanatory enough for this, people can just copy-paste the number also but it's obvious that's not the point of the demonstration

csfieldguide/chapters/content/en/human-computer-interaction/sections/interface-usability.md
courtneycb: I don't know if it will still be perfect when my PR is merged in.

eAlasdair: The important part is this sentence:

> For example, if you have a phone number to type in that someone has just told you, and it takes 12 seconds before you can type it, you may forget the number, whereas if you can access the interface within a couple of seconds, you can probably enter the number without much effort.

This comes after your interactive so it still works

csfieldguide/static/interactives/number-memory/scss/number-memory.scss
```diff
@@ -0,0 +1,30 @@
+.number-memory {
```
courtneycb: The border around this makes it look super inconsistent with the rest of the in-page interactives. I'm not a huge fan of it in whole-page mode either but that could just be personal preference.

When I first saw it I immediately thought it was a link to a whole page interactive because when you hover over one of those links it has a similar appearance.

courtneycb: I don't mind it around the distraction text though. I think the distraction text _does_ need something to kind of separate it from the chapter text (like the border).

csfieldguide/static/interactives/number-memory/js/number-memory.js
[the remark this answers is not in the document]
JackMorganNZ: You can use [user-select](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) to make it harder for them.

Issue #1098, opened 2019-07-22
Add archive website to footer links


JackMorganNZ:

courtneycb:
Resolved via #1102

PR #1102, opened 2019-07-22
Add link to release archive in footer.

courtneycb:
Resolves issue #1098. Note that the archive site is only available in English so the link will only appear in the footer on the English version of the site.


csfieldguide/templates/base.html
```diff
@@ -200,6 +200,14 @@ <h1>
                           {% trans "Curriculum Guides" %}
                         </a>
                       </li>
+                      {% if LANGUAGE_CODE == "en" %}
+                        <!-- Archive site is currently only available in English -->
```
JackMorganNZ:
```suggestion
                        <!-- Archive site is only available in English -->
```

PR #1117, opened 2019-07-31
Improve Regular Expression Filter Interactive

### LINKED ISSUE #1020: Improve Word Filter interactive
- The file should be loaded asynchronously, so that it doesn't lock up the webpage
- The word list should be under the interactive, so that it fits more neatly
- Consider making it a whole-page interactive only if it still doesn't fit neatly

eAlasdair:
Resolves #1020

- Words are now listed underneath the regex input
- Only the first 5000 matches are displayed, the user can then press a button to load more - in sets of 1000


csfieldguide/static/interactives/regular-expression-filter/js/regular-expression-filter.js
```diff
@@ -25,32 +29,59 @@ function processRegularExpression() {
...
     }
   }
-  display.innerHTML = html.trim();
+  updateWordCount();
+  displayMoreWords(5000);
 };
 
+// Displays 1000 more words (unless val is specified) and moves the 'display more' button to the end of the list
+function displayMoreWords(val) {
+  var num = 1000;
```
courtneycb: I think we could probably load 5000 words each time. Then you wouldn't need to pass in any parameters and the user doesn't have to click 'load more' as often.

eAlasdair: I'll change it but there's no reason to click load more in the first place. The interactive is about pattern matching; if they want the user can just specify to match words further down the alphabet. Also the more words that are displayed the longer it takes to display more words - up to a second per thousand in chrome towards the end

Issue #1017, opened 2019-06-20
Glossary terms missing


eAlasdair:
Now that we can easily see them in content on the website, we need to go back through it and add glossary terms to important words in the content

eAlasdair:
## HELP WANTED

If anyone has noticed words in the guide that would be useful to have a glossary definition for, please let me know

eAlasdair:
Potential terms to add, terms I think should be added for 3.0 are marked in bold:

**Also not that some terms are duplicated in multiple chapters. If we want the chapters to be standalone, then we need to include glossary definitions in all of them**

- Introduction chapter:
  - [ ] Computer
  - [ ] Digital device
  - [ ] **Programming/program** (defined in algorithms)
  - [ ] Interface
  - [ ] Software
  - [ ] **Binary**
- Algorithms Chapter
  - [ ] **Algorithm** (added later, could be moved up)
  - [ ] Data
  - [ ] pseudo-code
  - [ ] **Programming language**
  - [ ] **Complexity**
  - [ ] **Big O notation**
  - [ ] **adage**
  - [ ] analogy
  - [ ] Correctness
  - [ ] key concept
  - [ ] **Linear Search** (searches/sorts etc for consistency)
  - [ ] **Binary Search**
  - [ ] **Selection Sort**
  - [ ] Pattern
  - [ ] **Insertion Sort**
  - [ ] **Pivot** (noun)
  - [ ] **Mergesort**
  - [ ]  Sequencing/sequence
  - [ ] Iteration
  - [ ] computation
  - [ ] Parallelism
  - [ ] Timsort
  - [ ] **Artificial intelligence**
  - [ ] **Computer Vision**
  - [ ] **Hash/hashing**
  - [ ] Search Tree
  - [ ] Computer memory
  - [ ] Web server
  - [ ] **Theoretical**
  - [ ] **Empirical**
  - [ ] Hardware
  - [ ] Operating System
  - [ ] **Environment**
  - [ ] **Tractability**
- Programming Languages:
  - [ ] **programming/program**
  - [ ] **programming language**
  - [ ] Hardware
  - [ ] machine code
  - [ ] Interpreter
  - [ ] Assembler
  - [ ] Compiler
  - [ ] system
  - [ ] loop
  - [ ] **syntax**
  - [ ] variable
  - [ ] conditional
  - [ ] GUI
  - [ ] Arithmetic
  - [ ] Computer chip
  - [ ] Transistor
  - [ ] **Binary**
[...]

PR #1134, opened 2019-08-26
Content fixes + glossary entries for Coding chapters

eAlasdair:
Contributes to #1017 

- [x] Coding - Introduction
- [x] Coding - Compression
- [x] Coding - Encryption
- [x] Coding - Error Control

### New Glossary entries:

- Cost
- Dictionary attack
- Hash function
- Public key cryptography
- Salt
- Packet

### Renamed glossary entries

- Known plaintext attack (concatenate "plain-text")
- Plaintext (concatenate "plain text")

### Notes:

The whole story! section in the compression chapter was out of date, so i cleaned it up with as few changes as possible.

### DIFF
--- a/csfieldguide/chapters/content/en/coding-encryption/sections/the-key-distribution-problem.md
+++ b/csfieldguide/chapters/content/en/coding-encryption/sections/the-key-distribution-problem.md
@@ -6,7 +6,7 @@
 
 When describing an encryption scenario, cryptographers often use the fictitious characters "Alice" and "Bob", with a message being sent from Alice to Bob (A to B).
 We always assume that someone is eavesdropping on the conversation (in fact, if you're using a wireless connection, it's trivial to pick up the transmissions between Alice and Bob as long as you're in reach of the wireless network that one of them is using).
-The fictitious name for the eavesdropper is usually Eve.
+The fictitious name for the eavesdropper is usually Eve (E).
 
 {image file-path="img/chapters/xkcd-protocol.png" alt="A xkcd comic on protocols" hover-text="Changing the names would be easier, but if you're not comfortable lying, try only making friends with people named Alice, Bob, Carol, etc." source="https://xkcd.com/1323/"}
 
@@ -16,10 +16,12 @@ There are several other characters used to describe activities around encryption
 
 If Alice is sending an encrypted message to Bob, this raises an interesting problem in encryption.
 The ciphertext itself can safely be sent across an "unsafe" network (one that Eve is listening on), but the key cannot.
-How can Alice get the key to Bob? Remember the key is the thing that tells Bob how to convert the ciphertext back to plaintext.
+How can Alice get the key to Bob?
+Remember the key is the thing that tells Bob how to convert the ciphertext back to plaintext.
 So Alice can’t include it in the encrypted message, because then Bob would be unable to access it.
 Alice can’t just include it as plaintext either, because then Eve will be able to get ahold of it and use it to decrypt any messages that come through using it.
-You might ask why Alice doesn’t just encrypt the key using a different encryption scheme, but then how will Bob know the new key? Alice would need to tell Bob the key that was used to encrypt it... and so on... this idea is definitely out!
+You might ask why Alice doesn’t just encrypt the key using a different encryption scheme, but then how will Bob know the new key?
+Alice would need to tell Bob the key that was used to encrypt it... and so on... this idea is definitely out!
 
 Remember that Alice and Bob might be in different countries, and can only communicate through the internet.
 This also rules out Alice simply passing Bob the key in person.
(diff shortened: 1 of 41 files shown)


csfieldguide/chapters/content/en/glossary/substitution-cipher.md
```diff
@@ -1,6 +1,6 @@
 # Substitution cipher
 
 A type of cipher that works simply by replacing each letter or combination of letters in a plain text with a certain other letter or combination of letters to make up the cipher text.
```
courtneycb: I think this is a bit too long for a glossary entry. The first sentence by itself seems enough.

csfieldguide/chapters/content/en/coding-encryption/sections/the-key-distribution-problem.md
```diff
@@ -70,13 +72,13 @@ Some of the language may not be suitable for use in class, so discretion is need
 
 ## Public key systems
 
-One of the remarkable discoveries in computer science in the 1970s was a method called *public key encryption*, where it's fine to tell everyone what the key is to encrypt any messages, but you need a special private key to decrypt it.
+One of the remarkable discoveries in computer science in the 1970s was a method called {glossary-link term="public-key-cryptography"}public key encryption{glossary-link end}, where it's fine to tell everyone what the key is to encrypt any messages, but you need a special private key to decrypt it.
```
courtneycb: I don't think this needs a glossary tag here. The following sentence pretty much explains what it is, and the glossary link talks about public key cryptography in general, not just public key encryption.

csfieldguide/chapters/content/en/coding-error-control/sections/check-digits-on-barcodes.md
```diff
@@ -1,17 +1,17 @@
...
 These are all examples of situations that error control coding can help prevent.
 
 Barcode numbers, credit card numbers, bank account numbers, ISBNs, national health and social security numbers, shipping labels (serial shipping container codes, or SSCC) and tax numbers all have error control coding in them to help reduce the chance of errors.
-The last digit in each of these numbers is a check digit, which is obtained doing a special calculation on all the other digits in the number.
-If for example you enter your credit card number into a web form to buy something, it will calculate what the 16th digit should be, using the first 15 digits and the special calculation (there are 16 digits in a credit card number).
+The last digit in each of these numbers is a {glossary-link term="check-digit"}check digit{glossary-link end}, which is obtained by doing a special {glossary-link term="check-equation"}calculation{glossary-link end} on all the other digits in the number.
```
courtneycb: I don't think the glossary link for 'calculation' (check equation) provides any extra/useful info than what is already stated here. I suggest we don't put a glossary link for calculation here.

csfieldguide/chapters/content/en/coding-error-control/sections/whats-the-big-picture.md
```diff
@@ -1,24 +1,25 @@
...
+The {glossary-link term="bit"}bits{glossary-link end} on disks are very very small, and imperfections in the surface can eventually cause some of the storage to fail.
+The surfaces of CDs and DVDs are exposed, and can easily be damaged by storage (e.g. in heat or humidity) and handling (e.g. scratches or dust).
 Errors can also occur when numbers are typed in, such as entering a bank account number to make a payment into, or the number of a container that is being loaded onto a ship.
 A barcode on a product might be slightly scratched or have a black mark on it, or perhaps the package is bent or is unable to be read properly due to the scanner being waved too fast over it.
 Bits getting changed on permanent storage (such as hard drives, optical disks, and solid state drives) is sometimes referred to as data rot, and the [wikipedia page on bit rot](https://en.wikipedia.org/wiki/Data_degradation) has a list of more ways that these errors can occur.
 
-Nobody wants a computer that is unreliable and won’t do what it's supposed to do because of bits being changed!
+Nobody wants a computer that is unreliable and won’t do what it's supposed to do in a real-world scenario!
```
courtneycb: Unsure about this change. Reason?

eAlasdair: I didn't like "bits being changed" because it's too abstract - the focus is on the opinion of nobody(everybody) and very few people are going to think bits being changed is the reason for their frustration. However I think I got confused and replaced what I replaced it with with the real-world scenario bit so I'll think more about it

csfieldguide/chapters/content/en/coding-encryption/sections/the-key-distribution-problem.md
eAlasdair: They're the same thing: "public key" is important here, the rest is semantic. Cryptography is the Wikipedia version, encryption is used for the same thing in several other places (including here) so I could change a word to be consistent?

Also putting links in the sentence where it's explained is both consistent with what was done before I started, and appears to fit the [docs](https://cs-field-guide.readthedocs.io/en/latest/author/content_style_guide.html#content) (though ambiguous)

> Any jargon used needs to be clearly explained and should be considered as a glossary definition.

courtneycb: My main issue is that I think glossary definitions should be as concise as possible, and this one (in this particular context) is quite long. It feels like the information I get from the glossary definition is unecessarily interruptive seeing as the next sentence defines it just fine.

I think 'putting links in the sentence where it's explained' should be considered case by case.

Issue #1115, opened 2019-07-29
Rebuild 3D transform interactives.


courtneycb:
The archived interactives in the computer graphics chapter should be rebuilt for the Django system.

There are 2 main interactives, one with multiple versions:

- [x] Scene Editor
    - [x]  3D transform matrix 
    - [x]  3D translation matrix 
    - [x]  3D with multiple matrices and vectors
    - [x]  Scene creation

- [x] Matrix Simplifier

**TODO/suggestions**

- [x]  Need to look into adding copy funcitonality to matrix simplifier so we can copy matrices into scene editor.
- [x] Add remove all button to Matrix Simplifier

eAlasdair:
This is a superset of #374 so I'll close that one

eAlasdair:
And the remaining interactive is covered in #375

PR #1135, opened 2019-08-27
Rebuild matrix simplifier interactive for Django system

### LINKED ISSUE #375: Rewrite the "Matrix Simplifier" in the Graphics chapter
This needs to be updated: http://www.csfieldguide.org.nz/releases/1.9.9/_static/widgets/CG/CG-matrix-simplifier/CG-matrix-simplifier.html?info=Multiple%20transforms

courtneycb:
This is a complete rebuild of the old matrix simplifier from v1.9.9:

https://archive.csfieldguide.org.nz/1.9.9/_static/widgets/CG/CG-matrix-simplifier/CG-matrix-simplifier.html?info=Multiple%20transforms

Note that the old matrix simplifier (linked above) currently doesn't work so I do not know if I have included all of the old functionality.

The main functionality the new interactive has is the following:

- Ability for users to drag matrices and vectors to change their order in the calculation.
- Users can add their own matrices and vectors (has validation on the input boxes).
- Calculates the result of multiplying the matrices and adding the vectors, showing basic working and a simplified final matrix.

Relates to #1115 and resolves #375 

**Notes**

- The interactive uses MathJax to typeset the matrices. When we dynamically add matrices/vectors the math will not automatically be typeset (we must tell MathJax to typeset the new math). To prevent the flicker of non typeset math showing before Mathjax gets a chance to typeset it, we hide the relevant elements and only show them once it has been typeset. This still appears a little janky, but is much better than what it was. More info on typesetting here http://docs.mathjax.org/en/latest/advanced/typeset.html

- Users can only drag and drop matrices in the matrix container and vectors in the vectors container. This is because we multiply all of the matrices first and then add the vectors afterwards (scaling and rotation is applied to the point first, then translation), so it makes sense to keep them separate. They are also different dimensions (matrices are 3x3 while the vectors are 3x1) so vectors *must* be added last (not in between mutliplications of 3x3 matrices).

- ~~Mathjs treats arguments to trig functions as being in radians. Still need to figure out if this can be configured to degrees or if we need a work around (graphics chapter refers to degrees).~~ RESOLVED

- There is currently no limit on the number of matrices/vectors the user can add. Maybe there should be? Interested to hear thoughts on this.

- I don't know who the author is of the original matrix simplifier

**TODO**

- [x]  Add licences
- [x]  Add README
- [x]  Make sure chapter text matches interactive
- [x]  Figure out how to deal with Mathjs treating arguments to trig functions as being in radians
- [x]  Tidy up code
- [x]  Fix bug where if we remove the first matrix or vector and then add one, the added matrix/vector is not typeset (bug is larger than just this case but giving an example)

### DIFF
--- a/csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
+++ b/csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
@@ -6,12 +6,21 @@ const vsprintf = require('sprintf-js').vsprintf;
 const ROW_TEMPLATE = "%s & %s & %s";
 const MATRIX_TEMPLATE = "\\begin{bmatrix} %s \\\\ %s \\\\ %s \\end{bmatrix}";
 
-var matrices = [];
-var vectors = [];
+m1 = mathjs.matrix([[1,0,0],[0,1,0],[0,0,1]]);
+m2 = mathjs.matrix([[3,0,0],[0,3,0],[0,0,3]]);
+m3 = mathjs.matrix([[1,0,0],[0,1,0],[0,0,1]]);
+
+v1 = mathjs.matrix([[1], [0], [0]]);
+
+var matrices = [m1, m2, m3];
+var vectors = [v1];
+
 
 $(document).ready(function() {
   $('#add-matrix-from-input').click(addMatrix);
   $('#add-vector-from-input').click(addVector);
+  console.log(matrices);
+  console.log(vectors);
 });
 
 
@@ -23,6 +32,7 @@ function addMatrix() {
   $('#matrix-1').html(matrixString);
   // Request re-render. Taken from https://stackoverflow.com/questions/32239378/using-mathjax-in-an-updating-sequence-in-javascript
   MathJax.Hub.Queue(["Typeset", MathJax.Hub, 'matrix-1']);
+  showOutput()
 }
 
 
@@ -60,16 +70,17 @@ function formatMatrix(matrix) {
 
 function addVector() {
   vectorArray = [
-    $('#vector-row-0').val(), 
-    $('#vector-row-1').val(), 
-    $('#vector-row-2').val()
+    [$('#vector-row-0').val()], 
+    [$('#vector-row-1').val()], 
+    [$('#vector-row-2').val()]
   ];
   vector = mathjs.matrix(vectorArray);
   vectors.push(vector);
   vectorString = sprintf(MATRIX_TEMPLATE, vectorArray[0], vectorArray[1], vectorArray[2]);
-  $('#matrix-2').html(vectorString);
+  $('#vector-2').html(vectorString);
   // Request re-render. Taken from https://stackoverflow.com/questions/32239378/using-mathjax-in-an-updating-sequence-in-javascript
-  MathJax.Hub.Queue(["Typeset", MathJax.Hub, 'matrix-2']);
+  MathJax.Hub.Queue(["Typeset", MathJax.Hub, 'vector-2']);
+  showOutput()
 }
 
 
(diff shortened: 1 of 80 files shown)


csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
```diff
@@ -0,0 +1,543 @@
...
+var matrices = [m1, m2];
+var vectors = [v1];
+
+// Arrays that will keep track of the order the matrices and vectors are in
+var currentMatricesOrder = [m1, m2];
+var currentVectorsOrder = [v1];
+
+
+// only show equations once they are rendered
+var mjaxURL  = 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=TeX-AMS-MML_HTMLorMML,Safe.js';
```
eAlasdair: This line might need more explaining

csfieldguide/templates/interactives/matrix-simplifier.html
```diff
@@ -0,0 +1,190 @@
...
+        <button id="add-vector-btn" type="button" class="btn btn-primary d-none" data-toggle="modal" data-target="#vector-modal">{% trans "Add vector" %}</button>
+      </div>
+    </div>
+
+    <div class="row">
+      <h5 class="col-12 text-center">{% trans "Result" %}</h5>
+    </div>
+
+    <div class="row">
+      <div id="output-container" class="col-12 text-center invisible">
```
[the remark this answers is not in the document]
courtneycb: > Could these stay centered on the page but be left-aligned respective to each other?

Yeah they can, the only thing to note is the difference on _some_ mobile phones:

left-aligned:
image


centered:
image

Aside from this instance, I prefer left aligned as generally we should have the equal signs in line. Just bringing this up in case the mobile format changes things.

csfieldguide/templates/interactives/matrix-simplifier.html
```diff
@@ -0,0 +1,190 @@
...
+                \begin{bmatrix}
+                10 \\
+                0 \\
+                0 \\
+                \end{bmatrix}
+              \]
+            </div>
+          </div>
+        </div>
+        <button id="add-vector-btn" type="button" class="btn btn-primary d-none" data-toggle="modal" data-target="#vector-modal">{% trans "Add vector" %}</button>
```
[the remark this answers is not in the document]
courtneycb: Could we maybe make an issue to address this for later? I think this might reqiure a bit of time to implement.

csfieldguide/templates/interactives/matrix-simplifier.html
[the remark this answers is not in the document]
courtneycb: I would be happier to implement this as part of the transform interactive rebuild so it can be tested correctly :) would you be ok with that? Good idea.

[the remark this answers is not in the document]
eAlasdair: That's fine, for now I think just make sure there is at least a way to access the result so you don't have to rebuild everything later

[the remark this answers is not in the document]
eAlasdair: Could you force the lines to scroll rather than wrap?

csfieldguide/templates/interactives/matrix-simplifier.html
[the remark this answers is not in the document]
courtneycb: Issue here #1137

csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
```diff
@@ -0,0 +1,493 @@
...
+
+
+/**
+ * Displays the output of calculations
+ */
+function showOutput() {
+  // If output container is showing, hide it while mathjax renders.
+  if (!$('#output-container').hasClass('invisible')) {
+    $('#output-container').addClass('invisible');
+  }
```
eAlasdair: It doesn't matter if the element already has the class, so you don't need the if statement

csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
```diff
@@ -0,0 +1,493 @@
...
+    mathjs.number(inputEvaluated);
+    success = true;
+  }
+  catch {
+    $(this).addClass('input-error');
+    $('.add-from-input').prop('disabled', true);
+  }
+  if (success && $(this).hasClass('input-error')) {
+    $(this).removeClass('input-error');
+  }
```
eAlasdair: Similarly as before, this could be simplified

csfieldguide/templates/interactives/matrix-simplifier.html
```diff
@@ -0,0 +1,173 @@
...
+
+{% load i18n %}
+{% load static %}
+
+{% block html %}
+  <div class="container">
+    <h3 class='text-center'>{% trans "Matrix Simplifier" %}</h3>
+    <p class="text-center">
+      {% blocktrans %}
+      This interactive combines multiple matrix transformations to just one matrix that can apply each transformation with just one operation.<br>
```
eAlasdair: No need for the second 'just'

csfieldguide/templates/interactives/matrix-simplifier.html
```diff
@@ -0,0 +1,173 @@
+{% extends interactive_mode_template %}
```
eAlasdair: Now rapidly removing matrices results in a bunch of stuff being displayed for a couple frames:
out


Still:
image

It isn't a big deal so not worth putting much into, but perhaps a 'remove all' might be useful?

eAlasdair: image

A big enough number will push the matrix out of it's container juuust a teeny tiny bit

eAlasdair: Is the lack of background colour normal in the screenshot icons? (this is from the project panel 3D transforms)
It's also missing the matrices themselves but I guess that's a bit tricky
image

eAlasdair: Looking more closely this is pretty normal for our interactives - just most of the more obvious ones never turn up on different coloured panels

courtneycb: Have made an issue for this #1138

PR #1146, opened 2019-09-19
Rebuild scene editor interactive

courtneycb:
Handing this over to @eAlasdair as I have a higher priority task to do now.

What I've done so far:

- Configured the threejs scene using a skybox, lighting, cameras, textures, orbit controls etc.
- Added configuration for 4 different modes: transform, translation, multiple and scene-creation via url parameters.
- Added x,y,z axes (pos and neg)
- User can select different objects to add to the scene, these are currently added to the scene in a random position. Currently only 3 basic shapes are implemented: sphere, cube and cone. I would expect these to change as the chapter talks about adding a teapot (which should be the starting object - not a sphere which it currently is) and cups.
- Added skeleton code for retrieving matrix and vector entries.
- The licence file for the skybox images is under scene-editor/img - this should probably be moved to the third-party-licence file? Not sure where it fits.

### Alasdair hijacks the description here

Relates to #1115 

It has been decided that this will be a Minimum Viable Product type release. All the main functionality has been implemented, but more specific details will take significantly longer to implement.

## TODO

- Add teapot and cup meshes as described in the chapter content
- Add mesh for the point the camera orbits around (as practice for the teapot & cup ones)
- Consider adding more error control for user input
- Change the mesh materials so they appear more 3D
- Add limits to the matrices applied to objects

## Checks

- [x] Original base functionality implemented (as far as I can remember)
- [x] Properly licensed
- [x] properly documented


csfieldguide/static/interactives/scene-editor/js/scene-editor.js
```diff
@@ -0,0 +1,694 @@
...
+  // Axes
+  addAxes(size);
+
+  window.addEventListener( 'resize', onWindowResize, false );
+}
+
+/**
+ * Rescales the scene to the size of the window
+ */
+function onWindowResize() {
```
courtneycb: This probably doesn't matter for an MVP, but i'll put a note anyway. The scene doesn't stay in proportion when the window is resized, e.g the sphere appears stretched:

image

csfieldguide/templates/interactives/scene-editor.html
```diff
@@ -0,0 +1,99 @@
...
+          <div class="form-group">
+            <div class="input-group my-2">
+              <div class="input-group-prepend">
+                <label for="colour-input" id="colour-input-label" class="input-group-text">#</label>
+              </div>
+              <input type="text" maxlength="6" class="form-control" aria-describedby="random-colour-help" id="colour-input" placeholder="{% trans 'Colour (Hexadecimal)' %}">
+            </div>
+            <small id="random-colour-help" class="form-text text-muted">{% trans "Leave blank for a random colour" %}</small>
+          </div>
+          <button id="add-object" class="col-12 btn btn-primary" type="button">{% trans "Add object" %}</button>
```
courtneycb: Would be nice if we could also remove objects. Do you think this is needed for MVP though?

eAlasdair: Not going to be an easy addition. I'll add it to #1154

csfieldguide/static/interactives/scene-editor/js/scene-editor.js
eAlasdair: Hmm, that's unique to chrome

PR #1149, opened 2019-09-30
Amend content licence to exclude certain properties we don't own

eAlasdair:
Amends the LICENCE-CONTENT file to exclude:

- Images located within the `csfieldguide/static/img/chapters/` directory that are prefixed with `xkcd`
- `csfieldguide/static/img/third-party/`
- `csfieldguide/static/img/logos/`

as they may be subject to different licences


eAlasdair:
Tim is happy with the changes, so requesting @courtneycb's review just to get it released

PR #1151, opened 2019-10-03
Fix consistency in RSA interactives

eAlasdair:
Implements a fix I was meant to do in #1123 

The `e` component of keys is now split into pairs of characters (bytes) for consistency


courtneycb:
Did you ever try Jack's suggestion of using i-frames?

eAlasdair:
> Did you ever try Jack's suggestion of using i-frames? I missed checking that in the original PR. Aside from that I'm happy to approve your code, but would be good to not have the code duplication.

I did not, There ended up being several small differences between them. I can probably still do it but it's not as cut-and-dry as when I first asked the question

PR #1159, opened 2019-10-11
Add ability to edit existing equations in matrix simplifier interactive.

### LINKED ISSUE #1137: Add ability to edit matrices in matrix simplifier interactive
We should be able to edit existing matrices/vectors in the matrix simplifier interactive instead of having to remove an existing one and add a new one.

courtneycb:
This PR adds the ability to edit existing matrices and vectors in the matrix simplifier interactive.
Resolves #1137


csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
```diff
@@ -51,13 +52,32 @@ m1 = mathjs.matrix([
...
+  ['0'],
+  ['0']
+]
+// Arrays that will hold the string version of matrices and vectors.
+// Used for populating modal when editing a matrix or vector.
+// Needed because we don't want to display evaluated trig functions.
+// E.g we want to show cos(45) and not 0.71
+var matricesStringFormat = [m1String, m2String];
+var vectorsStringFormat = [v1String];
+
```
eAlasdair: (Nowhere in particular)
If there are no transform matrices the result should probably be the identity rather than 0
image
You can do this by starting with the identity matrix and multiplying each one in the list with that

csfieldguide/static/interactives/matrix-simplifier/scss/matrix-simplifier.scss
```diff
@@ -43,15 +43,30 @@
...
 }
 
+.close {
+  color: #7f7f7f;
+}
+
+.close:focus, .close:active {
+  outline: none !important;
+  box-shadow: none;
+}
```
eAlasdair: The x is darker than the pencil when you mouse over them, perhaps try to make them have the same hover colour

csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
```diff
@@ -72,7 +92,8 @@ $.getScript(mjaxURL, function() {
 $(document).ready(function() {
   $('#add-matrix-from-input').click(addMatrix);
   $('#add-vector-from-input').click(addVector);
```
eAlasdair: Not related to the changes I think but imo the default new vector should be 0, 0, 0 so (like the new matrix) by default it doesn't change anything
image

PR #1168, opened 2019-10-16
Matrix simplifier improvements

### LINKED ISSUE #1115: Rebuild 3D transform interactives.

courtneycb:
- Adds ability for resulting equation in `matrix-simplifier` to be copied and pasted into all versions of the `scene-editor` interactive.

- Adds ability to remove all equations in the `matrix-simplifier` interactive with one button click.

Resolves #1115


csfieldguide/static/interactives/scene-editor/js/scene-editor.js
```diff
@@ -124,13 +124,31 @@ $(document).ready(function () {
   $("#apply-transformation").click(applyTransformation);
   $("#delete-button").click(deleteSuspect);
 
-  $('.matrix-row input').on('keyup bind cut copy paste', validateInput);
-  $('.vector-row input').on('keyup bind cut copy paste', validateInput);
+  $('.matrix-row input').on('keyup bind cut copy', function() {
+    validateInput($(this));
+  });
+  $('.vector-row input').on('keyup bind cut copy', function() {
```
eAlasdair: Consider setting these to on `change` because it's not validating all of the input boxes on a paste event

Alternatively, at the end of your `populateInputsFromPaste` function, add:
```
  // Run the error-check
  $('.matrix-row input').trigger('keyup');
  $('.vector-row input').trigger('keyup');
```

I think i'd do the latter

csfieldguide/templates/interactives/matrix-simplifier.html
```diff
@@ -106,6 +106,22 @@ <h5 class="col-12 text-center">{% trans "Result" %}</h5>
...
+    </div>
+
+    <div class="row justify-content-center mb-5">
+      <button id="copy-eqtn" type="button" class="btn btn-primary col-md-2 col-sm-6 d-none" data-toggle="tooltip" data-placement="bottom">{% trans "Copy equation" %}</button>
+    </div>
+
+    <div class="row justify-content-center">
+      <button id="remove-all" type="button" class="btn btn-danger btn-sm col-md-2 col-sm-6 d-none">{% trans "Remove all" %}</button>
+    </div>
+
```
eAlasdair: (nowhere in particular)

I'm not a fan of the layout here
image
- What 'remove all' does is ambiguous, consider having two: one next to each 'add matrix/vector' button
- 'Copy equation' should specify that it'll be put on the clipboard i.e. 'copy to clipboard'
- Consider reducing the height of the code box to one line and using a scrollbar
- Consider: 'For convenience this code can be pasted into the Scene Editor interactive'

csfieldguide/static/interactives/scene-editor/js/scene-editor.js
```diff
@@ -902,3 +920,36 @@ function createTinyaxisMesh() {
   var object = new THREE.Mesh( geometry, material );
   return object;
 }
+
+
+/** Called on a paste event. Handles populating inputs copied from matrix simplifier 
+ *  or simply validates pasted input if it appears it hasn't been copied from
+ *  matrix simplifier.
+ */
+function populateInputsFromPaste(inputDiv) {
```
eAlasdair: We need to let the user know they can paste that MS code into any one of the boxes. Consider adding some text-muted instruction

csfieldguide/templates/interactives/matrix-simplifier.html
```diff
@@ -106,6 +106,22 @@ <h5 class="col-12 text-center">{% trans "Result" %}</h5>
...
+    <div class="row justify-content-center mb-1">
+      <small class="text-muted text-center">{% trans "The format below can be copied for pasting into the scene editor interactive." %}</small>
+    </div>
+
+    <div class="row justify-content-center mb-2">
+      <textarea id="code-to-copy" class="form-control col-md-4 col-sm-12" readonly></textarea>
+    </div>
+
+    <div class="row justify-content-center mb-5">
+      <button id="copy-eqtn" type="button" class="btn btn-primary col-md-2 col-sm-6 d-none" data-toggle="tooltip" data-placement="bottom">{% trans "Copy equation" %}</button>
```
eAlasdair: There is no tooltip for before the button is clicked. the RSA interactive equivalents have an extra `title` part for this

csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
```diff
@@ -363,6 +392,10 @@ function showOutput() {
   var matrixRows = matrixToArray(matrix);
   var vectorRows = matrixToArray(vector);
 
+  // update global result variable
+  resultEqtn = matrixRows.toString() + ',v,' + vectorRows.toString();
```
eAlasdair: Consider a leading 'm' for the code, then when parsing the pasted content you could check for the components individually i.e. If string includes `v,` then what follows is the vector, if string includes `m,` then what follows is the matrix

My reasoning is that sometimes users may only want to copy over the matrix part of the code, as they have the vector set already or something

csfieldguide/static/interactives/scene-editor/js/scene-editor.js
courtneycb: This isn't true when we allow them to copy only the matrix or only the vector part.
I added some text to try clarify but happy to hear suggestions.

Also I should probably add some sort of instruction on MS about how to only copy the matrix part or only the vector part, but unsure how to phrase it nicely right now.

csfieldguide/static/interactives/scene-editor/js/scene-editor.js
```diff
@@ -902,3 +920,36 @@ function createTinyaxisMesh() {
...
+    $('#matrix-row-0-col-1').val(matrixData[1]);
+    $('#matrix-row-0-col-2').val(matrixData[2]);
+
+    $('#matrix-row-1-col-0').val(matrixData[3]);
+    $('#matrix-row-1-col-1').val(matrixData[4]);
+    $('#matrix-row-1-col-2').val(matrixData[5]);
+
+    $('#matrix-row-2-col-0').val(matrixData[6]);
+    $('#matrix-row-2-col-1').val(matrixData[7]);
+    $('#matrix-row-2-col-2').val(matrixData[8]);
```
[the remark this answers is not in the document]
courtneycb: I did wonder.. still not sure why. 

Also with my changes since your review, the above input will actually put `1,2,3` in the vector. I don't think I can validate it perfectly anyways and is probably too much trouble for what it's worth.

csfieldguide/static/interactives/scene-editor/js/scene-editor.js
eAlasdair: My thinking is just explain that the user can paste it all (i.e. the copy to clipboard 'all') into any box.
The people who want to will try different things themselves; I'm saying it's good to account for other likely inputs but that doesn't mean we have to 'officially' support them

csfieldguide/static/interactives/scene-editor/js/scene-editor.js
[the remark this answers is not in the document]
eAlasdair: So long as we can account for both "m,x,x,x..." and "v,x,x,x" (plus the standard code) I think it's ok

csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
```diff
@@ -640,3 +680,25 @@ function validateInput() {
...
   }
 }
+
+
+/** Removes all equations and re-calculates output */
+function removeAllMatrices() {
+  $(".matrix").parent().remove();
+  currentMatricesOrder = [];
+  matricesStringFormat = [];
+  $('#remove-all-matrices').attr('disabled', true);
```
eAlasdair: Removing the last matrix manually doesn't disable the button
There is no adverse affect to this though
image

csfieldguide/templates/interactives/scene-editor.html
```diff
@@ -48,6 +48,9 @@ <h5 id="scene-creation-title" class="col-4 offset-2 heading-line-text">{% trans
             </div>
           </div>
         </div>
+        <div class="row justify-content-center my-1">
+          <small class="col-sm-8 offset-1 text-muted text-left pl-0">{% trans "You can paste in your code from Matrix Simplifier into any box!" %}</small>
```
eAlasdair: Take out the 'in', the sentence reads better without it

PR #1170, opened 2019-10-17
Update sentence about mesh points in CG chapter.

courtneycb:
We now use an entirely different teapot mesh in the `scene-editor` interactive.~~The new mesh has 289 points instead of 3644. Have updated the sentence to match.~~

We do not know how many points there are exactly now. Unless someone knows how to calculate how many there are?


eAlasdair:
I got the value of 289 from counting the lines [here](https://github.com/uccser/cs-field-guide/blob/develop/csfieldguide/static/js/third-party/threejs/TeapotBufferGeometry.js#L108). However, since there are a lot of circles in the teapot, there's a chance those circles are defined as position:radius instead, in which case the resulting mesh would have many more actual vertices

Now I suggest saying "many thousands" instead of defining a number

PR #1187, opened 2019-11-07
Use default settings for split health checks.

courtneycb:
We are currently having issues where we cannot set a custom path under `liveness_check` or `readiness_check` (across all projects) and this is causing deployments to fail. We will revert to using the default split health check settings until we know more about the issue.

Info on default settings here: https://cloud.google.com/appengine/docs/flexible/custom-runtimes/migrating-to-split-health-checks#understanding_key_differences

> By default split health checks only confirm that the VM instance and the Docker container are running. As long as these conditions hold, the VM will continue to receive traffic and remain alive regardless of the app's internal state.


eAlasdair:
Does this include updating the dev site health checks? They are still the outdated ones

courtneycb:
> Does this include updating the dev site health checks? They are still the outdated ones

I have updated the health checks on the dev site using:
`gcloud app update --split-health-checks --project cs-field-guide-dev`

courtneycb:
Codecov failed but got the approval from Jack to merge anyway.

PR #1193, opened 2019-11-18
Add suggested improvements to Data Bias interactive.

### LINKED ISSUE #1190: Suggested improvements for Data Bias interactive
- Fix line spacing
- On first level, fade to grey instead of snapping
- Try Black or White for initial background - so it's not immediately obvious there's a trick going on
   - If white, use a border
- Combinations of the above

courtneycb:
- [x] Fix line spacing (thanks @eAlasdair)
- [x] On first level, fade to grey instead of snapping
- [x] Try Black or White for initial background - so it's not immediately obvious there's a trick going on
- [x] Use a border
- [x] Potentially alter wording - waiting on @caitlind 

Decision yet to be made on whether we start with black or white or a random colour.
The coloured box still jumps around on mobile at the moment.

Resolves #1190


caitlind:
I've added my changes to the text, opinions on what I've written?

> First stage: "Forced perspective like this can influence what data you collect, and can bias overall results."
> Slider stage: "Changing your perspective can decrease data bias."

I had to change to formatting a tiny bit on the second stage so that the text didn't shift the display downwards. Does that still look ok to everyone?

Oops, just realised I left the old text in there commented out, I'll go delete that!

csfieldguide/static/interactives/data-bias/js/data-bias.js
```diff
@@ -3,8 +3,10 @@ const wNumb = require('wnumb');
 
 const COLOURS = ['red', 'lime', 'blue', 'yellow', 'purple', 'darkorange', 'fuchsia', 'deepskyblue'];
 const START_TEXT = gettext("Click each dot that you see on the screen, then click 'Next stage' to reveal the answer!");
-const MISSED_CIRCLES_TEXT = gettext('You seem to have missed some dots! <br><br>Forced perspective like this can be used in data representation and cause bias in the overall results.');
-const SLIDER_TEXT = gettext('Click and drag the slider to change the background colour. <br>What do you notice is happening?')
+// const MISSED_CIRCLES_TEXT = gettext('You seem to have missed some dots! <br><br>Forced perspective like this can be used in data representation and cause bias in the overall results.');
+// const SLIDER_TEXT = gettext('Click and drag the slider to change the background colour. <br>What do you notice is happening?')
+const MISSED_CIRCLES_TEXT = gettext('You seem to have missed some dots! <br><br>Forced perspective like this can influence what data you collect, and can bias overall results.');
+const SLIDER_TEXT = gettext('Click and drag the slider to change the background colour. Changing your perspective can decrease data bias.')
```
eAlasdair: > You seem to have missed some dots! Forced perspective like this can influence what data you collect, and can bias overall results.

I spent several minutes commenting on the original passage :cry: 
This version is a lot better. I don't have any gripes

> Click and drag the slider to change the background colour. Changing your perspective can decrease data bias.

There is absolutely nothing wrong with this statement; changing your perspective _can_ reduce data bias. so feel free to ignore anything after this full stop. But I still don't really like it... sorry! Maybe make it a bit more general like 'how you see data can affect how you interpret it' &ndash; it's not just about reducing bias, it's also about knowing how bias can _increase_ under certain conditions.
Or just "finding the right perspective" instead of 'changing your perspective'

eAlasdair:
> I had to change to formatting a tiny bit on the second stage so that the text didn't shift the display downwards. Does that still look ok to everyone?

Looks fine, I don't think what you described should have happened though - all the text sits in a box big enough for all of it

csfieldguide/static/interactives/data-bias/js/data-bias.js
caitlind: > But I still don't really like it... sorry! 

All good! I was getting stuck on that one as well and wasn't entirely happy with it.

> Maybe make it a bit more general like 'how you see data can affect how you interpret it' – it's not just about reducing bias, it's also about knowing how bias can increase under certain conditions.

I really like that, good point about how it is about seeing how bias can happen as well as reducing it. I'll change it to this (and hopefully the line spacing doesn't get messy!)

>  "finding the right perspective" instead of 'changing your perspective'

This one I'm not so sure about, since in many situations there probably won't be a "right" perspective that can be agreed upon, or it'll always be subjective in some way.

eAlasdair: :+1: If the line spacing breaks mess around with [these two values](https://github.com/uccser/cs-field-guide/pull/1193/files#diff-e96fe59a41a4347ab2ec3507be5697e6R127)

caitlind: Changed the text now, I just made one small change to your suggestion and switched "how you **see** data" to "how you **view** data"

Issue #1197, opened 2019-11-21
Minor issues in the Computer Graphics chapter


caitlind:
I have some suggestions for few minor changes I think could be made to the graphics chapter to make it more understandable:

- [ ] In section 12.2 Before the block interactives it would be good to explicitly say what directions negative and positive numbers will move the box in.

- [ ] The first interactive is quite confusing when you start using it, because of how the camera moves and how it is difficult to understand whether you are moving or if the block is moving. 
I'm not sure how or if this actually needs to be changed in any way (since it already explains in the description that the camera follows the box) but if anyone has any ideas it could be worth exploring them! Maybe adding visible axes?

- [ ] In section 12.2 starting from the "Swapping coordinates" interactive the idea of how the image is being rotated *around the x and y axes* needs to be explained more clearly. When I first read it I thought the word "rotated" had been used incorrectly and it should have been "mirrored".
I think some of this part needs re-writing and maybe should be divided into "rotation" and then "swapping coordinates. The interactive chould have 3 phases - rotating around x axis only, rotating around y axis only, and then both. 
![image](https://user-images.githubusercontent.com/18223869/69289680-31cbf580-0c62-11ea-9dfa-65daba43c0ef.png)

- [ ] In section 12.2.3 3D Transforms the word "vector" is used without being defined or explained. In this context it is about using vectors for translation (not vector graphics). The concept of a vector should probably be explained here.

eAlasdair:
> In section 12.2 starting from the "Swapping coordinates" interactive the idea of how the image is being rotated around the x and y axes needs to be explained more clearly. When I first read it I thought the word "rotated" had been used incorrectly and it should have been "mirrored".
I think some of this part needs re-writing and maybe should be divided into "rotation" and then "swapping coordinates. The interactive chould have 3 phases - rotating around x axis only, rotating around y axis only, and then both.

It's not even both, it's a y-axis rotation (i.e mirror/flip), then a z-axis rotation. It's definitely not a "simple rotation transformation"

eAlasdair:
> In section 12.2 starting from the "Swapping coordinates" interactive the idea of how the image is being rotated around the x and y axes needs to be explained more clearly. When I first read it I thought the word "rotated" had been used incorrectly and it should have been "mirrored".
I think some of this part needs re-writing and maybe should be divided into "rotation" and then "swapping coordinates. The interactive chould have 3 phases - rotating around x axis only, rotating around y axis only, and then both.

[...]

PR #1200, opened 2019-11-25
Resolve some issues with CG chapter

eAlasdair:
Addresses some of the issues raised in #1197 

## Changes

> In section 12.2 Before the block interactives it would be good to explicitly say what directions negative and positive numbers will move the box in.
- [x]  Add, in list form
> The content mentions point (3, 1) as the rightmost corner of the shape. This refers to the original arrow. The rightmost points of the current shape are (2, 3) and (2, 1)
- [x] Change to the upper rightmost point, (2, 3) and fix relevant calculations

### DIFF
--- a/csfieldguide/chapters/content/en/computer-graphics/sections/graphics-transformations.md
+++ b/csfieldguide/chapters/content/en/computer-graphics/sections/graphics-transformations.md
@@ -11,7 +11,7 @@ You can move it around using what's called a {glossary-link term="translation"}t
 - A positive or negative change in *y* will shift the box *up or down* respectively.
 - A positive or negative change in *z* will shift the box *in or out* respectively.
 
-The camera will always focus on the cube, so to see the back of the cube move it behind the camera with a positive *z* translation.
+**The camera will always focus on the cube**, so to see the back of the cube move it behind the camera with a positive *z* translation.
 Try typing 3D coordinates into this interactive to find each symbol.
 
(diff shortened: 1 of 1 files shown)


csfieldguide/chapters/content/en/computer-graphics/sections/graphics-transformations.md
```diff
@@ -5,8 +5,13 @@ In fact, every pixel you see in an image has usually had many calculations made
...
 The following interactive shows a cube with symbols on each face.
-You can move it around using what's called a {glossary-link term="translation"}translation{glossary-link end}, which simply adjusts where it is placed in space.
-The camera will always focus on the cube, so to see the back of the cube move it behind the camera.
+You can move it around using what's called a {glossary-link term="translation"}translation{glossary-link end}, which simply adjusts where it is placed in space:
+
+- A positive or negative change in *x* will shift the box *right or left* respectively.
+- A positive or negative change in *y* will shift the box *up or down* respectively.
+- A positive or negative change in *z* will shift the box *in or out* respectively.
+
+The camera will always focus on the cube, so to see the back of the cube move it behind the camera with a positive *z* translation.
```
caitlind: I think these changes also sufficiently address the second bullet point in the issue:

> The first interactive is quite confusing when you start using it, because of how the camera moves ...

The only thing I'd suggest is just putting "always focus on the cube" in bold text. I think the new text that's been added here will make the interactive make more sense, so the interactive probably wont need any changes

PR #1207, opened 2019-12-05
Fix overflow of matrices in Matrix Simplifier

### LINKED ISSUE #1138: Deal with overflow of matrices in matrix simplifier interactive
image

Note that we cannot simply round the numbers as they must remain as strings for the display of trig functions. We also cannot limit the number of characters users input as for it to not overflow the limit would have to be 10 characters and this is too short.

We tried adding scrollbars but it was rather inconsistent (see below)

recording-34-6

eAlasdair:
- [x] Adds a 'handle' for the user to drag the matrices around, instead of the whole boxes
- [x] Sets the matrices to scroll instead of overflow their boxes

I chose a + symbol for the handle icon because that is what was used in the Dragula examples. The hover and grab mouse symbols have been set appropriately but users still might confuse it for an 'add' button, so I changed it to an = symbol

Resolves #1138


csfieldguide/static/interactives/matrix-simplifier/js/matrix-simplifier.js
```diff
@@ -186,7 +188,9 @@ function appendInput(type, inputHtml) {
   var $closeButton = $('<button type="button" class="close dismiss-eqtn" aria-label="Close">');
   $closeButton.append($('<span aria-hidden="true">&times;</span>'));
   var $editButton = $('<button type="button" class="close edit-eqtn" aria-label="Edit">');
+  var $moveLabel = $('<span class="close move-eqtn handle" aria-label="Move">=</span>');
```
courtneycb: I think the equals sign looks too similar to a hamburger menu. Perhaps pick an icon from here? https://icons8.com/icons/set/drag

PR #1255, opened 2020-04-22
Set crossorigin attribute on the default image in pixel viewer

courtneycb:
Inconsistencies have been occurring on the pixel viewer interactive regarding CORS requests. After some investigating we noticed:

- Upon disabling cache through dev tools we no longer got a CORS error.
- Two requests for images were being made; one of which was being made through jquery.
- In the response headers there is an attribute `sec-fetch-mode` that has the value `no-cors`.
- Only the default image was triggering the CORS error.

We suspect that the jquery request is being cached first (which doesn't have CORS headers), and then the second request is made (with CORS headers) but during this request Chrome fetches from the cache (but the CORS headers aren't cached due to the first request) which results in a CORS error on the second fetch.

With this change we see only one request being made and it appears to have the `sec-fetch-mode` attribute set to `cors` locally.


courtneycb:
Closing as we are creating a release branch from develop.

PR #1263, opened 2020-05-05
Allow users to choose the number of cards displayed in binary cards interactive with input box

### LINKED ISSUE #1262: Allow number of cards to be changed for binary cards interactive
Add a dropdown at the bottom of the interactive to allow the user to change the number of cards that are displayed. Also a URL parameter should be added that disables this selector, which would be useful for specific examples we use in the text.

courtneycb:
**Changes**

- Add input box to the binary cards interactive that allows the user to change how many cards are displayed. This can be anywhere from 1 to 16 cards inclusive.

- Add `input` URL parameter. Can be set to either 'true' or 'false to show or hide the input box respectively. Default is set to 'true'.

**Notes**

- The interactive now loads 16 (the max) cards first, before hiding cards to show the default 8 cards (or the value of the `digits` URL parameter if it is passed). Mostly done to prevent rewriting majority of the code. This allows us to easily hide and show cards dynamically when the user interacts with the input box.

- If a `start` parameter is passed that is less than 16 (or whatever the max is set to) letters long, e.g `BBBBB` we add padding to the start of the string so that the correct cards are flipped to the correct side. E.g `BBBBB` would become `WWWWWWWWWWWBBBBB`. This is not shown in the URL.

- When the numbers get large they overflow the cards but this currently happens on prod. I have created an issue for this.

Resolves #1262

### DIFF
--- a/csfieldguide/static/interactives/binary-cards/README.md
+++ b/csfieldguide/static/interactives/binary-cards/README.md
@@ -10,16 +10,16 @@ This interactive is created to teach binary numbers, and emulates the Binary Car
 The interactive shows the cards with dots from **128** to **1** by default (128, 64, 32, 16, 8, 4, 2, 1), however these can be configured by the following parameters:
 
 - `base=value` - Where `value` is the number base to use (defaults as 2).
-- `digits=value` - Where `value` is the amount of digits to display (default is 8). Min is 1 and max is 16.
-- `offset=value` - Where `value` is the amount to offset the displayed digits (default is 0). Using a positive number will show the placings from the `digits` + `offset` value, for the number of given digits. For example, using a base of `10`, digits as `3`, and offset as `2` will show the 100,000, 10,000, 1,000, and 100 placings. Using a negative number for the `value` will display floating numbers.
-- `start=sides` - Where `sides` is a sequence of `W` and `B` characters, to state the sides that should be displayed when the interactive starts. The first letter states whether the first card (on the left) should be white (`W`) or black (`B`), the second letter stands for the second card. Therefore the number of letters should match the number of digits used.
+- `cards=value` - Where `value` is the number of cards to display (default is 8). Min is 1 and max is 16.
+- `offset=value` - Where `value` is the amount to offset the displayed cards (default is 0). Using a positive number will show the placings from the `cards` + `offset` value, for the number of given cards. For example, using a base of `10`, cards as `3`, and offset as `2` will show the 100,000, 10,000, 1,000, and 100 placings. Using a negative number for the `value` will display floating numbers.
+- `start=sides` - Where `sides` is a sequence of `W` and `B` characters, to state the sides that should be displayed when the interactive starts. The first letter states whether the first card (on the left) should be white (`W`) or black (`B`), the second letter stands for the second card. Therefore the number of letters should match the number of cards used.
 - `input=value` - Where `value` is `true` or `false` (default is true). Indicates whether or not to display the input box that lets users choose how many cards are displayed.
 
 ### Examples
 
-- `?base=16&digits=4` - Shows 6 (default) cards for the first 4 digits of base 16 (hexadecimal).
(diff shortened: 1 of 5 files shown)


eAlasdair:
Maybe let the user type in a number they want, then limit it to [1, 16]

csfieldguide/templates/interactives/binary-cards.html
```diff
@@ -13,11 +13,15 @@ <h3 class="text-center mt-2">{% trans "Binary Cards" %}</h3>
...
-            {% trans "Flip all card sides to:" %}
+          <div class="mb-2">
+            <p class="d-inline mr-2">{% trans "Flip all card sides to:" %}</p>
             <button id="flip-to-black" class="btn btn-dark">{% trans "Black" %}</button>
             <button id="flip-to-white" class="btn btn-light">{% trans "White" %}</button>
-          </p>
+          </div>
+          <div id="cards-input" class="form-group form-inline">
+            <label for="cards-to-show" class="mr-2">{% trans "Choose how many cards to show:" %}</label>
+            <input id="cards-to-show" class="form-control" type="number" min="1" max="16" value="8"/>
```
eAlasdair, line 23: I think the box, and possibly the text description, should be shorter

csfieldguide/static/interactives/binary-cards/README.md
```diff
@@ -9,9 +10,10 @@ This interactive is created to teach binary numbers, and emulates the Binary Car
...
 
 - `base=value` - Where `value` is the number base to use (defaults as 2).
-- `digits=value` - Where `value` is the amounts of digits to display (default is 6).
+- `digits=value` - Where `value` is the amount of digits to display (default is 8). Min is 1 and max is 16.
 - `offset=value` - Where `value` is the amount to offset the displayed digits (default is 0). Using a positive number will show the placings from the `digits` + `offset` value, for the number of given digits. For example, using a base of `10`, digits as `3`, and offset as `2` will show the 100,000, 10,000, 1,000, and 100 placings. Using a negative number for the `value` will display floating numbers.
 - `start=sides` - Where `sides` is a sequence of `W` and `B` characters, to state the sides that should be displayed when the interactive starts. The first letter states whether the first card (on the left) should be white (`W`) or black (`B`), the second letter stands for the second card. Therefore the number of letters should match the number of digits used.
```
eAlasdair, line 15: What should happen if the string for the `start` parameter is longer than the number of digits? (or vice versa)
i.e. `?digits=2&start=BBBBB`
At the moment the `digits` parameter takes priority, so only 2 cards get shown, and I think that could actually be useful.
Likewise if the `start` parameter is shorter than the number of digits, `digits` still takes priority, and 'W' cards are (from the user's perspective) prepended to the cards defined.
Perhaps this should be explained in the README

csfieldguide/static/interactives/binary-cards/README.md
```diff
@@ -9,9 +10,10 @@ This interactive is created to teach binary numbers, and emulates the Binary Car
 The interactive shows the cards with dots from **128** to **1** by default (128, 64, 32, 16, 8, 4, 2, 1), however these can be configured by the following parameters:
 
 - `base=value` - Where `value` is the number base to use (defaults as 2).
-- `digits=value` - Where `value` is the amounts of digits to display (default is 6).
+- `digits=value` - Where `value` is the amount of digits to display (default is 8). Min is 1 and max is 16.
```
eAlasdair, line 13: Change to `number` rather than `amount`
An aside, but I still don't understand why it's called digits. I guess they're 'binary digits' but since that's the only reference to binary in the interactive, 'cards' seems more appropriate

csfieldguide/static/interactives/binary-cards/js/binary-cards.js
```diff
@@ -26,18 +38,57 @@ $(document).ready(function () {
         updateDotCount();
     });
 
+    // Update number of cards shown based on input
+    $('input').change(function() {
```
eAlasdair, line 42: Should text input default to 1?
Maybe if input is negative try the absolute value first

Issue #1253, opened 2020-04-21
UI issues with LZSS compression interactive


JackMorganNZ:
- Fix bug where spacing between sets of boxes looks like a space, when spaces are within boxes.
- Investigate adding arrows, and mode to show all arrows at once.

eAlasdair:
I wouldn't call the first bullet point a bug, It was designed to show groupings of characters: on the one hand the space between groups is ambiguous but on the other contracting the space would make it more difficult to see the groupings themselves.

Perhaps instead we should have a visible space symbol like this short bucket type one
![image](https://user-images.githubusercontent.com/25916475/79939314-e9d02980-84b2-11ea-9ee5-b6d9df8d46df.png)

courtneycb:
We should investigate making the highlights permanent on click since mobile users can't hover.

eAlasdair:
Other suggestions:
- [ ] Write the encoded form out as text
- [ ] Show a comparison of the text size before/after encoding

PR #1273, opened 2020-05-27
Fix issue with LZSS compression interactive encoding text incorrectly

### LINKED ISSUE #1271: LZSS compression interactive not encoding text correctly
For the example below it should replace all but the first 2 characters with a pointer.

image

courtneycb:
This PR addresses a bug in the LZSS interactive where the text is not being encoded correctly for some strings.

**The problem** 
The sliding window and encoded message are initialised to the first 6 characters. This means the first 6 characters can never be pointers - even if there are repeating strings within these 6 characters. This does not work correctly for small strings or for strings of a repeating nature, e.g `ababababababab`.

**Changes**

- Initialise sliding window and encoded message to only contain the first 2 characters (minimum match length).
- Raise max length of match to 16 (4 bits). This used to be 5 which is unusually small.
- Replace space characters in the encoded message with the open box character. This addresses part of #1253
- Add small note under the encoded text stating that spaces are replaced with the open box character.

Resolves #1271

### DIFF
--- a/csfieldguide/chapters/content/en/coding-compression/sections/general-purpose.md
+++ b/csfieldguide/chapters/content/en/coding-compression/sections/general-purpose.md
@@ -13,7 +13,7 @@ The interactive below illustrates this idea using a variation of the Ziv-Lempel 
 
 In LZSS, a reference to where some text occurred before is actually two numbers: the first says how many characters to count back to where the previous phrase starts, and the second says how long the referenced phrase is.
 Each reference typically takes the space of about one or two characters, so the system makes a saving as long as two characters are replaced.
-The options in the interactive above allow you to require the replaced length to be at least two characters, to avoid replacing a single character with a reference.
+The interactive above requires the replaced length to be at least two characters, this avoids replacing a single character with a reference.
 Of course, all characters count, not just letters of the alphabet, so the system can also refer back to the spaces between words.
 In fact, some of the most common sequences are things like a full stop followed by a space.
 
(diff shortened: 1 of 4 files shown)


csfieldguide/templates/interactives/lzss-compression.html
```diff
@@ -21,8 +21,8 @@ <h3>{% trans "LZSS Compression" %}</h3>
         </div>
         <div id="lzss-compression-right-box">
           <p>{% trans "Encoded Message" %}:</p>
-          <div id="lzss-compression-compressed-text">
-          </div>
+          <div id="lzss-compression-compressed-text"></div>
+          <p id="char-replacement-note">{% trans "Note: spaces have been replaced with the &#9251; character." %}</p>
```
eAlasdair, line 25: Maybe it'd be better to say spaces _are_ replaced with the bucket character.

eAlasdair:
Also this sentence in the chapter content:
> The options in the interactive above allow you to require the replaced length to be at least two characters, to avoid replacing a single character with a reference.

image

There are no replaced length options for the interactive, so please reword this sentence

csfieldguide/chapters/content/en/coding-compression/sections/general-purpose.md
```diff
@@ -13,7 +13,7 @@ The interactive below illustrates this idea using a variation of the Ziv-Lempel
 
 In LZSS, a reference to where some text occurred before is actually two numbers: the first says how many characters to count back to where the previous phrase starts, and the second says how long the referenced phrase is.
 Each reference typically takes the space of about one or two characters, so the system makes a saving as long as two characters are replaced.
-The options in the interactive above allow you to require the replaced length to be at least two characters, to avoid replacing a single character with a reference.
+The interactive above requires the replaced length to be at least two characters, this avoids replacing a single character with a reference.
```
eAlasdair, line 16: Actually do we need this sentence at all? It's self explanatory and the sentence is now a repeat of the one previous

PR #1285, opened 2020-06-29
Add text output and efficiency calculation to LZSS interactive

eAlasdair:
Related to #1253 
- Add encoded message output in textual form
- Add display of unencoded message size and approximate encoded message size
- Functionality to include the number of reference flags in the encoding size calculation exists, but commented out as I thought it would confuse users
image


courtneycb:
Has Tim looked at this?

eAlasdair:
> Has Tim looked at this?

Yes we discussed it

csfieldguide/static/interactives/lzss-compression/js/lzss-compression.js
```diff
@@ -33,15 +39,64 @@ function newLineDiv() {
...
+    var compressed_text_div = document.getElementById('message-to-decode');
+    var text_to_write = "";
+    var item;
+
+    for (var i=0; i < encoded_message.length; i++) {
+        item = encoded_message[i];
+
+        if (item == newlineCharacter) {
+            text_to_write += "\n";
+        } else if (item.length == 1) { // just a single character)
```
courtneycb, line 81:
```suggestion
        } else if (item.length == 1) { // just a single character
```

csfieldguide/static/interactives/lzss-compression/js/lzss-compression.js
```diff
@@ -33,15 +39,64 @@ function newLineDiv() {
...
 
     // create new div for first line of the message
     var line_div = newLineDiv();
 
     var index = 0;
-    for (var i = 0; i < encoded_message.length; i++) {
+    for (var i=0; i < encoded_message.length; i++) {
         var string = encoded_message[i];
         
         if (string == newlineCharacter) {
```
courtneycb, line 102: on line 111

`if (string.length == 1)` will trigger a TypeError for newline characters because the string is undefined

courtneycb: If I add a bunch of newlines after some text it will not update the text representation of the encoded message

image

courtneycb:
Not related to this PR but is there a reason why the lzssAlgorithm js file is camel case?

eAlasdair:
> Not related to this PR but is there a reason why the lzssAlgorithm.js file is camel case?

Because heapsAlgorithm.js (city trip) is camelcase, but I created that file too so... no

csfieldguide/static/interactives/lzss-compression/js/lzss-compression.js
eAlasdair: Regarding your first comment, there are two issues:
1) the newline character is `:n` so 2 characters, not undefined
2) the `continue` statement means that part of the code wont execute for newlines anyway (this is code from the original so it does work)

However the problem itself is happening, nice spot

eAlasdair: Wait line 111 is for the visual representation and it succeeds hmmm

PR #1286, opened 2020-06-29
Increase Binary Cards card limit

### LINKED ISSUE #1284: Reimplement arbitrary number of cards in Binary Cards interactive
It was reduced to a maximum of 16 for various reasons, but we've been asked to put it back

Also we should re-add the `digits` parameter for backwards compatibility

eAlasdair:
Resolves #1284 

The number can't easily be made arbitrary but after resolving a bit of logic any maximum number up to >1024 can be specified. I chose 256 as it's approaching the limit of a reasonable time to execute balanced against the usefulness of having so many cards

EDIT: To resolve visual problems, rebranded to the maximum of 16 or the value of the `cards` parameter. So it can now be arbitrary, but the input box cannot be set any higher

Also:
- Reimplemented the `digits` url parameter for backwards compatibility. `cards` takes preference
- Made an example in the readme slightly clearer


csfieldguide/static/interactives/binary-cards/js/binary-cards.js
```diff
@@ -1,13 +1,14 @@
 var urlParameters = require('../../../js/third-party/url-parameters.js')
 "use strict";
-MAX_NUM_CARDS = 16;
+MAX_NUM_CARDS = 256;
 DEFAULT_NUM_CARDS_TO_SHOW = 8;
```
courtneycb, line 5: (nowhere in particular) We seem to have lost dots for some cards and the total number of dots value is wrong. This is with the query string base=10

image

csfieldguide/templates/interactives/binary-cards.html
```diff
@@ -20,7 +20,7 @@ <h5 id="dot-decimal-count"></h5>
           </div>
           <div id="cards-input" class="form-group form-inline">
             <label for="cards-to-show" class="mr-2">{% trans "Cards to show:" %}</label>
-            <input id="cards-to-show" class="form-control col-md-1 col-4" type="number" min="1" max="16" value="8"/>
+            <input id="cards-to-show" class="form-control col-md-1 col-4" type="number" min="1" value="8"/>
```
courtneycb, line 23: Should we have the max set to 256 here too?

csfieldguide/static/interactives/binary-cards/js/binary-cards.js
eAlasdair: :scream: I didn't check

csfieldguide/templates/interactives/binary-cards.html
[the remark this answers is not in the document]
eAlasdair: No, it's set at the start of the JS using the MAX_NUM_CARDS const

csfieldguide/static/interactives/binary-cards/js/binary-cards.js
courtneycb: Yay for reviews!

PR #1291, opened 2020-07-02
Add ability to show colour codes in Hexadecimal on Pixel Viewer

### LINKED ISSUE #1277: Add Hexadecimal colour code option to Pixel Viewer

courtneycb:
**Changes**

- Users now have 3 different options for how colour codes should be displayed.
    - Decimal (as separate RGB values) **(default)**
    - Hexadecimal (as separate RGB values)
    - Hexadecimal (web colour)

- Add optional URL parameter `colour-code` that sets what colour code the interactive loads with. Values are:
    - `rgb` (default)
    - `rgb-hex`
    - `hex`

- Update README.

Resolves #1277

### DIFF
--- a/csfieldguide/static/interactives/pixel-viewer/js/pixel-viewer.js
+++ b/csfieldguide/static/interactives/pixel-viewer/js/pixel-viewer.js
@@ -48,6 +48,8 @@ this.images = [
 this.tiling = new Tiling;
 this.piccache = Array();
 
+var show_hex = true
+
 const image_base_path = base_static_path + 'interactives/pixel-viewer/img/';
 var source_canvas = document.getElementById('pixel-viewer-interactive-source-canvas');
 
@@ -872,6 +874,10 @@ $('#pixel-viewer-interactive-show-pixel-fill').change(function() {
     scroller.finishPullToRefresh();
 });
 
+$("input[name='colourCode']").click(function() {
+    var colour_code = $("input[name='colourCode']:checked").val()
+})
+
 // Caches data about the image
 function init_cache(width, height){
   piccache = Array()
@@ -987,7 +993,6 @@ var paint = function(row, col, left, top, width, height, zoom) {
 
     // If text opacity is greater than 0, then display RGB values
     if (text_opacity > 0) {
-        context.font = (14 * zoom).toFixed(2) + 'px Consolas, Courier New, monospace';
         if (!show_pixel_fill) {
             context.fillStyle = "rgba(0, 0, 0, " + text_opacity + ")";
         } else if ((((pixelData[0] / 255) + (pixelData[1] / 255) + (pixelData[2] / 255)) / 3) < 0.85) {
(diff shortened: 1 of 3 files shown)


csfieldguide/static/interactives/pixel-viewer/js/pixel-viewer.js
```diff
@@ -996,14 +1019,34 @@ var paint = function(row, col, left, top, width, height, zoom) {
...
 };
 
+// Taken from https://stackoverflow.com/questions/5623838/rgb-to-hex-and-hex-to-rgb
+function componentToHex(c) {
+  var hex = c.toString(16);
+  return hex.length == 1 ? "0" + hex : hex;
+}
+
+function rgbToHex(r, g, b) {
+  return "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
```
eAlasdair, line 1047: Looks like Tim was wanting individual R G and B components like before
> The default in both cases should still be decimal, but they need an option to show the numbers 0..255 as 00..FF. At this stage I don't think they should have '0x' in front of them.

So I guess that'd mean an option between decimal and hexadecimal - it makes more sense too

csfieldguide/templates/interactives/pixel-viewer.html
```diff
@@ -39,6 +39,21 @@ <h5>{% trans "Options" %}</h5>
                     <input id="pixel-viewer-interactive-show-pixel-fill" type="checkbox" checked="checked">
                     {% trans "Show pixel background" %}
                 </label>
+                <div class="mb-1">
+                    {% trans "Colour code: " %}
+                    <div class="form-check form-check-inline">
+                        <input class="form-check-input" type="radio" name="colourCode" id="rgb-colour-code" value="rgb" checked>
+                        <label class="form-check-label" for="rgb-colour-code">{% trans "Decimal (as separate RGB values)" %}</label>
```
eAlasdair, line 46: maybe just `(separate RGB)` for the two for concision. I think they can be stacked to look a bit nicer
image

csfieldguide/static/interactives/pixel-viewer/js/pixel-viewer.js
```diff
@@ -996,14 +1017,39 @@ var paint = function(row, col, left, top, width, height, zoom) {
...
+          }
+        }
     }
 };
 
+// Taken from https://stackoverflow.com/questions/5623838/rgb-to-hex-and-hex-to-rgb
+function componentToHex(c) {
+  var hex = c.toString(16);
+  return hex.length == 1 ? "0" + hex : hex;
+}
```
eAlasdair, line 1047: I think the hex should be uppercase to have consistent-height digits, but it might get confusing if B (blue) is misinterpreted as B (11)

csfieldguide/static/interactives/pixel-viewer/js/pixel-viewer.js
```diff
@@ -50,6 +50,7 @@ this.piccache = Array();
 
 const image_base_path = base_static_path + 'interactives/pixel-viewer/img/';
 var source_canvas = document.getElementById('pixel-viewer-interactive-source-canvas');
+var colour_code_rep = 'rgb';
```
eAlasdair, line 53: (nowhere in particular) Funnily enough this doesn't matter too much, as the values will be the same if the squares are the same colour, but if it's easy could we get the hex numbers closer to their characters?
It looks like 00 R rather than R 00 etc
image

PR #1309, opened 2020-08-05
Fix up Pixel Viewer and RGB mixer plus modernise CMY mixer

### LINKED ISSUE #1306: Make CMY Mixer consistent with the RGB Mixer
In #1305 the colour code was added to the center of the coloured box in the RGB Mixer and some layout changes were made - particularly for smaller screen sizes.

We should make the same changes to the CMY Mixer.

eAlasdair:
- [x] Fix indentation style problems in PV
- [x] Fix style of RGB Mixer URL parameters
  - `pixelmania-2` redirects to a different url but the same functionality
  - The original `mode=pixelmania` option is deprecated but still works
  - New parameters: `hide-selector=true|false` for the rgb/hex radio options & `pixelmania` for the logo
- [x] Modernise CMY Mixer:
  - New URL parameter `hide-selector=true|false` for the rgb/hex radio options
    - No `pixelmania` parameter because it wasn't used for Pixelmania
  - Layout and css rules brought inline with RGB Mixer
  - Display of cmy colour code in colour box
     - There is no equivalent hex code for CMY so it is ~left blank~ also cmy colour code but using hex values

Resolves #1306

### DIFF
--- a/csfieldguide/static/interactives/cmy-mixer/js/cmy-mixer.js
+++ b/csfieldguide/static/interactives/cmy-mixer/js/cmy-mixer.js
@@ -134,14 +134,31 @@ CMY_Mixer.setColor = function(){
   } else {
     cmy_as_rgb = cmy_to_rgb(cyan_val, magenta_val, yellow_val);
   }
-	var color = 'rgb(' +
-		cmy_as_rgb[0] + ',' +
-		cmy_as_rgb[1] + ',' +
-		cmy_as_rgb[2] + ')';
 
-	// Fill the color box.
-	CMY_Mixer.result.style.background = color;
-	CMY_Mixer.result.style.color = color;
+  if (!useHex) {
+    // Display colour as text in coloured box
+    // There is no equivalent hex representation of colour in CMY
+    // _technically_ it should be CMYK since this is describing what printers do, but let's not get into that
+    var font_colour = getFontColour(cmy_as_rgb[0], cmy_as_rgb[1], cmy_as_rgb[2]);
+    var colour_text = 'cmy(' +
+      cyan_val + ',' +
+      magenta_val + ',' +
+      yellow_val + ')';
+    $('#interactive-cmy-mixer-colour-code')
+      .text(colour_text)
+      .css('color', font_colour);
+    } else {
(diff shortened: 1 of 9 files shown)


csfieldguide/static/interactives/cmy-mixer/js/cmy-mixer.js
```diff
@@ -158,11 +179,26 @@ CMY_Mixer.setColorFromInputBox = function() {
     yellow_val = parseInt(yellow_val, 16);
   }
   var cmy_as_rgb = cmy_to_rgb(cyan_val, magenta_val, yellow_val);
+  
+  if (!useHex) {
+    // Display colour as text in coloured box
+    // There is no equivalent hex representation of colour in CMY
```
courtneycb, line 185: Do you think we need to state this somewhere on the interactive? Not sure what the best thing to do is.

We could:
- Show the RGB hex code equivalent and explain it
- Show no hex code and explain why
- Leave it as is but then users may get confused as to why it doesn't show anything?

:woman_shrugging: :woman_shrugging: :woman_shrugging:

csfieldguide/static/interactives/cmy-mixer/scss/cmy-mixer.scss
```diff
@@ -39,24 +43,36 @@
...
-@media (max-width: 259px) {
-  #interactive-cmy-mixer-container {
-    flex-direction: column;
-    padding-right: 0rem;
-  }
-  #interactive-cmy-mixer-result {
-    width: 100%;
-    height: 5rem;
-    margin-bottom: 1rem;
+
```
courtneycb, line 46: Too many blank lines here

csfieldguide/static/interactives/cmy-mixer/js/cmy-mixer.js
eAlasdair: Well since cmy isn't actually used either and the string doesn't show any new information I'm a big fan of not having the string form at all. We should discuss it

courtneycb: I think I'd be a fan of not having the string form as well.

courtneycb (approved):

PR #1333, opened 2020-10-08
Improve limitations for Algorithm Timer interactive

### LINKED ISSUE #1332: Improve limitations for Algorithm Timer Interactive
image

These hard limits for the Algorithm Timer interactive are far too restrictive when the complexity isn't at the top end (i.e. O(n!)), and as a result you can't even use it to do [the exercises in the content](https://www.csfieldguide.org.nz/en/chapters/complexity-and-tractability/tractability/). It would be far more sensible to remove/loosen these limits on lower complexities (perhaps O(n^3) and below), and warning users if they attempt to do it at higher complexities.

eAlasdair:
Resolves #1332 

- [x] Limit the number of items `n` based on the algorithm complexity:
    - If the number of steps required (based on `n` & the complexity) is greater than `10^32`, the calculation won't be made
- [x] increase limit on the number of processors, and number of operations per second:
    - New limit is `10^16` processors/operations per second
- [x] Add second warning message for each input box
   - One for an invalid number, one for a number that's too high

### DIFF
--- a/csfieldguide/static/interactives/algorithm-timer/README.md
+++ b/csfieldguide/static/interactives/algorithm-timer/README.md
@@ -1,5 +1,13 @@
 # Algorithm Timer Interactive
 
 **Author:** Courtney Bracefield
+**Modified by:** Alasdair Smith
 
 This interactive is created to demonstrate the running time for algorithms of different complexity.
+
+## Required files
+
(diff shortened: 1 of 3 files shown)


eAlasdair:
One problem with the n input box is that it will do the calculation to figure out if the number is too high. If it's too high the thing will freeze

courtneycb (approved):
Great work!

PR #1340, opened 2020-10-27
Do not throw error in Big Number Calculator when the base of the exponent is 1

courtneycb:
EDIT:
After some investigation I ended up adding back in the old check and altering it further.
- If the exponent is 1 or 2 digits long the calculation terminates in reasonable time no matter how long the base is.
- If the exponent is 3 or more digits long then the x length + y length restriction is applied. I decided to test the highest base numbers possible for each digit length (999, 9999, 9999, ... etc) and find the highest exponent that could be entered with the calculation still terminating in ~5 seconds. Remembering that if the base has 3 digits then the exponent is limited to 6 digits because of the first restriction etc. I then used the results of that test to set a maximum value for the exponent depending on the number of digits in the base. This is in no way an optimal solution - for instance 2^150000 would terminate in reasonable time, but because 9^150000 doesn't the limit is set to 50000, so 2^150000 "can't" be calculated and so on... definitely open to suggestions for a better solution!

--------------------------------------------------------------------------------------------------------------------------------------------------------------------------

This issue was raised to us by a user.

If the base is 1 then the calculator still throws an error.
BNC

This can happen because:
1. The `pow` function in Big.js can not handle calculations where the exponent is greater than 1 million.
2. We check if the combined digit length of x and y is greater or equal to 10, if it is we throw an error.

We now check if the base (x) is equal to 1, if it is we just set the result to 1.

~~I removed the check for the combined digit length of x and y being greater or equal to 10 because:~~

~~1. If the exponent was 1 and the base had 10 or more digits an error would be thrown.
2. We display the error message if the number can't be calculated anyway. This means that if the number can be calculated it will be, else it will display the error message. I also didn't see a reason to have this restriction? Let me know if it's there for a reason.~~ 

Also included a word change requested by Tim.


eAlasdair:
> I removed the check for the combined digit length of x and y being greater or equal to 10 because:
> 1   If the exponent was 1 and the base had 10 or more digits an error would be thrown.
> 2   We display the error message if the number can't be calculated anyway. This means that if the number can be calculated it will be, else it will display the error message. I also didn't see a reason to have this restriction? Let me know if it's there for a reason.

The automatic restriction doesn't appear to work all that well, ie 9^999999 is allowed (both live and in this pr) yet I had to stop my browser after 5 minutes of it trying. Similarly 9999^999999 is rejected on the old BNC (due to 10 chars) but is allowed (and doesn't finish) in your PR with the manual restriction removed.

courtneycb:
> > I removed the check for the combined digit length of x and y being greater or equal to 10 because:
> > 1   If the exponent was 1 and the base had 10 or more digits an error would be thrown.
> > 2   We display the error message if the number can't be calculated anyway. This means that if the number can be calculated it will be, else it will display the error message. I also didn't see a reason to have this restriction? Let me know if it's there for a reason.
> 
> The automatic restriction doesn't appear to work all that well, ie 9^999999 is allowed (both live and in this pr) yet I had to stop my browser after 5 minutes of it trying. Similarly 9999^999999 is rejected on the old BNC (due to 10 chars) but is allowed (and doesn't finish) in your PR with the manual restriction removed.

Added an edit to my PR description to describe the changes I made!

eAlasdair (changes requested):
It's a bit jarring to be hitting the different calculation buttons and suddenly the thing freezes for a few seconds. Could you change the result box to `Calculating...` or something while it calculates instead of leaving the old result there?

eAlasdair:
Also don't let it queue more button presses! Disable the buttons or something

csfieldguide/static/interactives/big-number-calculator/css/big-number-calculator.scss
```diff
@@ -25,3 +25,8 @@
 #interactive-big-number-calculator textarea.materialize-textarea:focus:not([readonly]) {
   box-shadow: none !important;
 }
+.calculating {
```
eAlasdair, line 28:
```suggestion
#interactive-big-number-calculator div#interactive-big-number-calculator-result.calculating {
```

For consistency

eAlasdair (approved):

JackMorganNZ (approved):

PR #1366, opened 2021-01-31
Fix trainsylvania-complete uninteractive

eAlasdair:
Fixes the Trainsylvania Complete uninteractive, where it couldn't be shown in-page and was replaced by a static English image if the screen was too small.
It is now a constant size that can be displayed on small screens and the English image has been removed

This fix is included in #1203 but deserving of its own PR


courtneycb (approved):
Nice one!

PR #1369, opened 2021-02-11
Add percent-encoder to cfg-parsing-challenge

eAlasdair:
Adds a teacher-mode-only button that opens a modal which the user can use to set the grammar productions in the interactive.

Since there's no button to switch to teacher mode on the interactive page itself, the user needs to be in teacher mode before opening the interactive, or to switch to teacher mode in another tab and then refresh the interactive page, to see this button.

The instructions button toggles a dropdown of a few key bullet points about syntax, the example one drops down a representation of the default productions.

Once the grammar productions have been set, the user can obtain a link, adequately percent-encoded, to the interactive with the given productions. This link also hides the generator

The link generated can also leave the boundary of the modal, `word-wrap: break-word` doesn't help so not sure how to proceed - scrollbar?

**Edit**: *Also removed the ambiguity of edge cases in URL grammar productions regarding numbers. Whereas there was a shortcut for lone integers previously; all terminals must now be enclosed in inverted commas.*

image

image

### DIFF
--- a/csfieldguide/static/interactives/cfg-parsing-challenge/README.md
+++ b/csfieldguide/static/interactives/cfg-parsing-challenge/README.md
@@ -11,7 +11,8 @@ The interactive has the ability to do any grammar via use of the URL parameters.
 
 ### Basic Parameters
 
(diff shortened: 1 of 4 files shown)


JackMorganNZ (approved):

JackMorganNZ (approved):

eAlasdair:
Superseded by #1378

PR #1461, opened 2021-10-05
Further changes to CFG interactive

### LINKED ISSUE #1420: Further changes to CFG Parser Challenge interactive
- Disable the text field that allows customising the equation text, as it's prone to errors.
  - Allow it to be enabled via a query parameter or button.
  - Test limiting inputs to only known terminals.
- Lower interface priority of 'New productions' button, hide it towards the bottom of the interactive as a additional feature.
- Possibly increase average difficulty of random equations but attempting to generate with a minimum number of productions.
- Toggle percent encoding of URL for readability.

eAlasdair:
Resolves #1420 

- [x] Lower interface priority of 'New productions' button, hide it towards the bottom of the interactive as a additional feature.
    - Now the button is labelled `Customise Productions`
    - Now in bottom right and coloured grey (**Note:** Can still be removed entirely with parameter `hide-builder`)
- [x] Disable the text field that allows customising the equation text, as it's prone to errors.
    - Can be re-enabled with URL parameter `editable-target`
- [x] Increase average difficulty of random equations.
    - They are still random (I didn't force a minimum number of productions) but the difficulty skew is now more closely proportional to the probability of following each production.
- [x] Use wording 'generate' instead of 'form'
- [x] Fix wrapping behaviour of user expression
- [x] Remove internal shortcut for storing terminals as a list of elements as it no longer makes sense
- [x] Prefill production builder with appropriate YACC syntax for the existing productions
- [x] Use an en-dash in spaces rather than a hyphen to make the increasing-integers-production shortcut clearer when viewing negative numbers (e.g. -9--1 becomes -9 &#8211; -1)
- [x] Update README

image

### DIFF
--- a/csfieldguide/static/interactives/cfg-parsing-challenge/README.md
+++ b/csfieldguide/static/interactives/cfg-parsing-challenge/README.md
@@ -11,7 +11,7 @@ The interactive can demonstrate most grammars via URL parameters.
 
 This interactive can be configured by appending various parameters to the default URL.
 To do this, add a question mark (`?`) after the last `/` in the URL, then each parameter separated by ampersands (`&`).
-For this interactive, each parameter is in the form `keyword=value`, where each `keyword` is an option to be changed and `value` is what that option should be set to.
+For this interactive, each parameter is in the form `keyword=value`, where `keyword` is an option to be changed and `value` is what that option should be set to.
 For example:
 
 - The URL `[...]/cfg-parsing-challenge/` could be changed to `[...]/cfg-parsing-challenge/?hide-generator=true&examples=1+1|2+3+4`
@@ -26,7 +26,7 @@ For example:
 Examples will be cycled through in the order given.
 - `productions=str`: Set the grammar productions that can be used.
 An example of the correct syntax using the default grammar can be found below.
-  - When setting your own productions, it is recommended that you also use either the `recursion-depth` or `hide-generator` parameters.
+  - When setting your own productions, it is recommended that you also use either the `recursion-depth` or `hide-generator` parameter.
 
 The built-in equation generator works by following productions at random from the initial non-terminal - essentially building a parse tree - until there are no non-terminals left in the equation.
 In certain grammars, the parse tree could continue indefinitely, so the generator is forced to stop a certain number of branches down the tree:
--- a/csfieldguide/static/interactives/cfg-parsing-challenge/README.md
+++ b/csfieldguide/static/interactives/cfg-parsing-challenge/README.md
@@ -1,6 +1,6 @@
 # CFG Parsing Challenge Interactive
 
-**Author:** Alasdair Smith
+**Author:** Alasdair Smith  
 **Modified by:** Jack Morgan, Alasdair Smith
 
 This interactive demonstrates a context-free grammar (CFG) by allowing a user to use it to build a mathematical equation (default behaviour) or sentences.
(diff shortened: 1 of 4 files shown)


eAlasdair:
> @eAlasdair Could you please write an explanation here of the new logic changes, in particular how `findPossibleExpressionsAtDepth()` works to populate `expressionsAtDepth_`, and then how that variable is used in `recursiveRandomExpression()`. I understand parts of it, but struggling to see how it all fits together. Having it written here will be good documentation for the next developer too.

The aim of this algorithm is to find everything that can be produced by following productions with a maximum depth, so that we don't need to bogosearch for valid expressions to use as examples.

The first note to make is that the phrase "expressions at depth" means "expressions that can be made with the given maximum depth". So `expressionsAtDepth_` is a store of all possible valid expressions **up to** a given depth, organised by that depth and the starting nonterminal. At depth 1, a valid expression can only be made if the nonterminal is replaced by a terminal, this is the first step of `findPossibleExpressionsAtDepth(depth)`, and for the default productions results in:
```
expressionsAtDepth_ = {
    1: {
        E: [],  // No production for E ends in a terminal string so the list is empty
        N: [['0'], ['1'], ['2'], ['3'], ['4'], ['5'], ['6'], ['7'], ['8'], ['9']],
    }
}
```
**Note 1**:  The inner dictionary follows the normal internal syntax, the inner lists are for separating terminals and nonterminals from eachother. e.g look at `const DEFAULT_PRODUCTIONS`
**Note 2**: I'm leaving the syntax for defining a string (`"`) out; the terminals in `expressionsAtDepth_` include the inverted commas

For every depth above 1, `findPossibleExpressionsAtDepth(depth)` looks at each production and figures out if it's possible to get from the nonterminal to a valid expression, using that production. To do that it needs to check that all nonterminals as a result of the production can themselves reach a valid expression at any depth lower than the current one. For example, at depth 2:
- N &#10142; 0|1|... are all valid productions because they end in a terminal string
- E &#10142; N is valid because we know (looking at `expressionsAtDepth_`) that at depth 1 N can get to a terminal string
- E &#10142; E + E is NOT valid because we know that at depth 1, E cannot get to a terminal string

**Note**: In the code, `innerSuccess` is `true` if *one* of the nonterminals in the sub-expression reaches a valid expression, `outerSuccess` is `true` if *all* of the nonterminals in the sub-expression reach a valid expression. 

Valid results are stored as expressions with pointers to other results, with the pointer being [depth, nonterminal] -> Follow the nonterminal at that depth to find the rest of the expression. The pointers are stored in a list as there could be multiple options, so the result for depth 2 is:
```
expressionsAtDepth_ = {
    1: { as above },
    2: {
        E: [
            [[[1, N]]] // A list in the usual syntax, of a list of pointers, of the pointer [depth=1, nonterminal=E]
        ],
        N: [['0'], ['1'], ['2'], ['3'], ['4'], ['5'], ['6'], ['7'], ['8'], ['9']],
    }
}
```

And for depth 3:
```
expressionsAtDepth_ = {
    1: { as above },
    2: { as above },
    3: {
        E: [
            // N
            [[[1, N], [2, N]]]  // <1N OR 2N>
            // E + E & E * E
            [[[2, E]],  '+', [[2, E]]],  // <2E> + <2E>
            [[[2, E]],  '*', [[2, E]]],  // We don't need to know what E can get to at depth 2, only that it can get to something
            // -E
            ['-', [[2, E]]],                            // - <2E>
            // (E)
            ['(', [[2, E]], ')'],                       // ( <2E> )
        ],
        N: [['0'], ['1'], ['2'], ['3'], ['4'], ['5'], ['6'], ['7'], ['8'], ['9']],
    }
}
```
**Note**: There are definitely ways to make this more spatially efficient: since N only goes to nonterminals we only need to store it at depth 1; and in theory if a depth can be found then every depth above it is guaranteed. But even as is, it's more spatially efficient than storing every possible permutation normally and far faster than any method we've so far come up with.

The result of the algorithm is a data structure we can follow at random without reaching a dead (invalid) end. If we want a random expression with maximum depth 3 we:

- `randomExpression(3)`
- Expressions have already been parsed
- `recursiveRandomExpression(3, E):
    - If the list for `expressionsAtDepth_[3][E]` is empty we know it's impossible to get to a valid expression and don't need to bother parsing the tree. It's a big problem if this isn't the first level of recursion because it implies that further up the tree is a path that says a valid expression can be gotten via this node, which is false.
    - we pick a sub-expression at random from `expressionsAtDepth_[3][E]`, let's say `['(', [[2, E]], ')']`
    - for each element in the sub-expression we either have a terminal string or a list of pointers
    - '(' is a terminal, so we can move it on to our return string as is
    - `[[2, E]]` is a list of pointers, we pick one at random, there's only one: `[2, E]`
    - `recursiveRandomExpression(2, E)`
        - pick a sub-expression at random: `[[[1, N]]]`
        - pick a random pointer: `[1, N]`
        - `recursiveRandomExpression(1, N)`
            - pick a sub-expression at random: `['1']`
            - the sub-expression contains only terminal strings, so we return it joined: `'1'`
        - we move `'1'` on to our return string
        - the loop is finished so we return our return string `'1'`
    - we move '1' on to our return string: `'(''1'`
    - we move ')' on to our return string: `'(''1'')'`
    - the loop is finished so we return our return string `'(''1'')'`
    - // The pairs of inverted commas (`''`) not being removed here is actually a minor bug, but it isn't needed - the end result is the same
- we remove the inverted commas, then return the result: `(1)`

And the new user target with depth 3 is `(1)`

PR #2011, opened 2022-11-16
Set UID of node user to the UID of the host user

Yamboy1:
## Proposed changes

On multi-user systems, such as the University of Canterbury linux systems, it appears that the node user created by the Nodejs docker base has a different UID to the host user. This is a problem because we use a bind mount, which retains the host users permissions, meaning that the node user doesn't have permission to access the mounted files.

This PR sets the UID of the node user to the value of the $UID bash shell variable to fix the permission issue.

### Checklist

*Change the space in the box to an `x` for those you have completed.
You can also fill these out after creating the pull request.
If you're unsure about any of them, don't hesitate to ask.
We're here to help!
This is simply a reminder of what we are going to look for before merging your change.*

- [x] I have read the contribution guidelines.
- [x] I have linked any relevant existing issues/suggestions in the description above (include `#???` in your description to reference an issue, where `???` is the issue number).
- [x] The pull request is requesting a merge into the `develop` branch.
- [x] I have added necessary documentation (if appropriate).
- [x] If I've added/modified/deleted a third-party system, I've updated the relevant license files.
- [x] I have added myself to the 'Community Contributors' section of `csfieldguide/templates/appendices/contributors.html`

### DIFF
--- a/.github/workflows/test-and-deploy.yaml
+++ b/.github/workflows/test-and-deploy.yaml
@@ -32,6 +32,8 @@ jobs:
         uses: actions/checkout@v3
       - name: Create Docker network
         run: docker network create uccser-development-stack
+      - name: Set DOCKER_UID variable
+        run: echo "DOCKER_UID=$(echo $UID)" >> $GITHUB_ENV
       - name: Start systems
         run: docker compose -f docker-compose.local.yml up -d
       - name: Create static files
--- a/.github/workflows/test-and-deploy.yaml
+++ b/.github/workflows/test-and-deploy.yaml
@@ -19,6 +19,9 @@ jobs:
         uses: actions/checkout@v3
       - name: Create Docker network
         run: docker network create uccser-development-stack
+        # Required for the node service
+      - name: Set DOCKER_UID variable
+        run: echo "DOCKER_UID=$(echo $UID)" >> $GITHUB_ENV
       - name: Start systems
         run: docker compose -f docker-compose.local.yml up -d
       - name: Run Django system check
@@ -32,6 +35,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Create Docker network
         run: docker network create uccser-development-stack
+        # Required for the node service
       - name: Set DOCKER_UID variable
         run: echo "DOCKER_UID=$(echo $UID)" >> $GITHUB_ENV
       - name: Start systems
@@ -127,6 +131,9 @@ jobs:
 
       - name: Create Docker network
         run: docker network create uccser-development-stack
+        # Required for the node service
+      - name: Set DOCKER_UID variable
+        run: echo "DOCKER_UID=$(echo $UID)" >> $GITHUB_ENV
 
       - name: Start system
         run: docker compose -f docker-compose.local.yml up -d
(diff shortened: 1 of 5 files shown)


JackMorganNZ (changes requested):
There are major whitespace changes that are unneeded, and these also make it difficult to see what has changed. It you could revert the whitespace/indentation changes, then I can review it again.

Yamboy1:
Thank you. I didn't realise my autoformatter was still on.

JackMorganNZ (approved):
Thanks for making this change! It will be needed in our other projects too.

Yamboy1:
There are errors in the tests when trying to deploy this, not going to merge until i fix those and get a working development deployment. See https://github.com/Yamboy1/cs-field-guide/actions/runs/3483296577/jobs/5826626533

Yamboy1:
The CI passed on my fork, this PR is good again.

.github/workflows/test-and-deploy.yaml
```diff
@@ -19,6 +19,9 @@ jobs:
         uses: actions/checkout@v3
       - name: Create Docker network
         run: docker network create uccser-development-stack
+        # Required for the node service
+      - name: Set DOCKER_UID variable
+        run: echo "DOCKER_UID=$(echo $UID)" >> $GITHUB_ENV
```
JackMorganNZ, line 24: What is the value appended to `$GITHUB_ENV`?

Yamboy1: That line is essentially the same as this line in `./dev`: https://github.com/uccser/cs-field-guide/pull/2011/files#diff-dc2babd290214832274e2012c4d8af5cb43bcd1e5f3a8aa7955d6143d37b407aR4. It is needed because some jobs call `docker compose up` manually, rather than through `./dev`

Yamboy1: Also, the reason we need to set the variable at all, is because $UID is a bash variable and is not exposed to docker, so I need to wrap it in a environment variable

Yamboy1: @JackMorganNZ

JackMorganNZ: Sorry, my question is why is it not just: `run: echo "DOCKER_UID=$(echo $UID)"`.

The appending at the end doesn't seem to add anything in our pipeline.

Yamboy1: Oh, that is just how you set environment variables inside a github actions job.

JackMorganNZ: But why does the environment variable need to be set?

Yamboy1: Because this line now depends on the current users UID, this line will fail if the environment variable is not set. I have no idea why the default parameter in the line above doesn't work, but it doesn't, so we need the env variable.

JackMorganNZ: I found the doc page that explained it: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

This allows it to be used by subsequent steps in the job.

JackMorganNZ (approved):

PR #2015, opened 2022-11-20
Add information about the list of panel types

Yamboy1:
Add documentation about the different types of panels available with custom styles for authors in CSFG. Previously, the only way to know would be to look through the code, or look at other sections on the website as examples.


Yamboy1:
Have explanation on not needing to use the specified types, but they have a default style

